PR #3, opened 2017-09-06
Add CSS Parser

dsager:
This class allows the parsing of CSS code from google and offers methods to identify characteristics of specific class selectors.

This will later on be used to examine individual nodes of the HTML.


ifosch:
Quite long PR ;)
Tests are failing... looks like something missing in the code or the tests... maybe this needs some mock for `ArticleJSON::Import`?

miguelbernadi:
+1. I see there's no way to distinguish the case where only one class matches and all classes match. Is it not needed?

ifosch:
+1

dsager:
@miguelbernadi 
No, this is actually not needed. For example: we just have to check if a certain node has bold formatting, at that point we don't care if it's also italic.

PR #5, opened 2017-09-06
Add class to handle XML nodes from Google doc

dsager:
This class wraps around a nokogiri XML node and offers some analytical methods.

### DIFF
--- a/lib/article_json/import/google_doc/html/node.rb
+++ b/lib/article_json/import/google_doc/html/node.rb
@@ -1,0 +1,44 @@
+module ArticleJSON
+  module Import
+    module GoogleDoc
+      module HTML
+        class Node
+          attr_reader :node
+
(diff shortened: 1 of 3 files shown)


christofdamian:
+1

lib/article_json/import/google_doc/html/node.rb
```diff
@@ -0,0 +1,44 @@
...
+
+          # Check if a node equals a certain text
+          # @param [String] text
+          # @return [Boolean]
+          def has_text?(text)
+            node.inner_text.strip.downcase == text.strip.downcase
+          end
+
+          # Check if the node is empty, i.e. not containing any text or nested
+          # nodes of any importance - empty spans as not important ;)
```
yaneq, line 27: *are

yaneq:
+1 nice!

PR #7, opened 2017-09-06
Refactor Google HTML Node class

dsager:
- Fix issue in #empty? related to images
- Re-order & memoize analyze methods


christofdamian:
+1 isn't this a bit over optimized?

dsager:
it's useful for the next step, when where I have to call the methods multiple times. just not sure what happens under the hood in nokogiri with xpath and such things, so better safe than sorry :)

nicolas-fricke:
+1

PR #8, opened 2017-09-06
Add support for paragraphs

dsager:
Paragraphs (as one might expect) include text and none of the other supported elements that are wrapped in a `<p>` tag.


nicolas-fricke:
+1

christofdamian:
+1

PR #10, opened 2017-09-07
Add text elements

dsager:
Text elements are the basic unit of content, paragraphs will be composed of multiple text elements...

### DIFF
--- a/spec/article_json/import/google_doc/text_element_spec.rb
+++ b/spec/article_json/import/google_doc/text_element_spec.rb
@@ -1,0 +1,162 @@
+describe ArticleJSON::Import::GoogleDoc::HTML::TextElement do
+  subject(:element) do
+    described_class.new(text_node: nokogiri_node, css_analyzer: css_analyzer)
+  end
+  let(:nokogiri_node) do
+    Nokogiri::XML.fragment(xml_fragment.strip).children.first
+  end
+  let(:css_analyzer) do
+    ArticleJSON::Import::GoogleDoc::HTML::CSSAnalyzer.new(
+      <<-css
+        .bold { font-weight: bold }
+        .italic { font-style: italic }
+        .bold-italic { font-weight: bold; font-style: italic }
+      css
+    )
+  end
+
+  describe '#content' do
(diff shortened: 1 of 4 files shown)


christofdamian:
+1

spec/article_json/import/google_doc/text_element_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    let(:xml_fragment) do
+      <<-html
+        <p><span class="bold">A </span><span><a href="https://devex.com">link</a></span><span class="italic"> and styling.</span></p>
+      html
+    end
+
+    it 'parses all text sub-nodes' do
+      puts subject.map(&:to_h)
```
nicolas-fricke: This looks like a debug left-over 😉

spec/article_json/import/google_doc/text_element_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      expect(subject.size).to eq 3
+
+      expect(subject[0].content).to eq 'A '
+      expect(subject[0].bold?).to be true
+      expect(subject[0].italic?).to be false
+      expect(subject[0].href).to be nil
+
+      expect(subject[1].content).to eq 'link'
```
nicolas-fricke, line 151: Since these are all small unit tests, I think you could have separate `it` cases for the different expectancies.

dsager: I thought about it but then decided to have them properly grouped in an `it` block... No big deal IMO.

dsager: Also, I wouldn't know how to do that without using `its`, which is no longer a part of rspec-core => 3.x:  https://gist.github.com/myronmarston/4503509

nicolas-fricke: Well, you would have to do it with regular `it` blocks.. I always like more atomic tests, but I know that not everyone agrees with that here :)

nicolas-fricke:
+1

PR #14, opened 2017-09-07
Add image element

dsager:
An image is a node wrapping an `img` tag and another node that contains the caption, treated as text elements.


towanda:
+1

ifosch:
+1

PR #16, opened 2017-09-07
Add heading element

dsager:
A heading is created from a `h1`, `h2`, `h3`, `h4` or `h5` tag.

### DIFF
--- a/lib/article_json/import/google_doc/html/heading_element.rb
+++ b/lib/article_json/import/google_doc/html/heading_element.rb
@@ -1,0 +1,43 @@
+module ArticleJSON
+  module Import
+    module GoogleDoc
+      module HTML
+        class HeadingElement
+          # @param [Nokogiri::XML::Node] node
+          def initialize(node:)
+            @node = node
+          end
+
+          # The raw text content of the heading, without any markup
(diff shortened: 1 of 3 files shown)


christofdamian:
+1

lib/article_json/import/google_doc/html/heading_element.rb
```diff
@@ -0,0 +1,43 @@
...
+          # Determine the level of the heading
+          # The level corresponds to the header tag, e.g. `<h3>` is level 3.
+          # @return [Integer]
+          def level
+            case @node.name
+              when 'h1' then 1
+              when 'h2' then 2
+              when 'h3' then 3
+              when 'h4' then 4
+              when 'h5' then 5
```
towanda, line 26: What about:

```ruby
@node.name.tr('h','').to_i
```

dsager: then `h42` would also work :)

towanda: and that would be a feature, right?

towanda:
+!

towanda:
+1

PR #18, opened 2017-09-08
Add text box element & improve text parsing

dsager:
- Add text box element
  A text box can contain headings and paragraphs, in the future they should also support lists.
- Ignore empty nodes when extracting text


miguelbernadi:
+1

christofdamian:
+1

PR #19, opened 2017-09-08
Treat highlights as text boxes

dsager:
They are basically the same, so now both `Highlight:` and `Textbox:` will start a text box.


towanda:
+1

miguelbernadi:
+1

christofdamian:
+1

PR #23, opened 2017-09-08
Add main parser class

dsager:
The parser loops over all nodes of the HTML document and generates the elements. The integration test takes an export of the reference document, parses it and compares the result with the expected output from another fixture.


christofdamian:
+1

lib/article_json/import/google_doc/html/parser.rb
```diff
@@ -0,0 +1,109 @@
...
+          def parse_current_node
+            case @current_node.type
+            when :heading then parse_heading
+            when :paragraph then parse_paragraph
+            when :list then parse_list
+            when :image then parse_image
+            when :text_box then parse_text_box
+            when :quote then parse_quote
+            when :hr, :empty, :unknown then nil
+            end
```
ifosch, line 47: Not to be critic, and not sure if that would be really rubyist, but I feel implementing the whole HTML standards might be a very long case. Might be interesting to consider something like a hash of kinds of node to functions? Maybe something different but cleaner than just a hash, for example, a hookable code architecture?
These are not strong suggestions, just ideas coming raw to my mind.

ifosch:
+1

lib/article_json/import/google_doc/html/parser.rb
```diff
@@ -0,0 +1,109 @@
...
+          # @return [Object]
+          def parse_current_node
+            case @current_node.type
+            when :heading then parse_heading
+            when :paragraph then parse_paragraph
+            when :list then parse_list
+            when :image then parse_image
+            when :text_box then parse_text_box
+            when :quote then parse_quote
+            when :hr, :empty, :unknown then nil
```
towanda, line 46: When I have to do method dispatching in ruby, I really miss [clojure multimethods](https://adambard.com/blog/structured-clojure-protocols-and-multimethods/)

lib/article_json/import/google_doc/html/parser.rb
```diff
@@ -0,0 +1,109 @@
...
+            nodes = []
+            until NodeAnalyzer.new(@body_enumerator.peek).hr?
+              nodes << @body_enumerator.next
+            end
+            QuoteElement.new(nodes: nodes, css_analyzer: @css_analyzer)
+          end
+
+          # @return [Boolean]
+          def body_has_more_nodes?
+            @body_enumerator.peek
```
towanda, line 100: I didn't know about `.peek`, nice.

towanda:
+1

lib/article_json/import/google_doc/html/parser.rb
dsager: It's not really HTML, it's the elements we are supporting (text box for example is something we made up)...

I thought about doing this via magic methods or a hash, but found this easiest to read and understand at a first glance...

christofdamian: I agree. There probably should be a registry of classes for the different elements which then have their own parse method and maybe detection too. 
Case statements tend to be a bad smell.

dsager: I actually intended to implement the parsing in the individual classes.
But the problem is that some methods need to move the "cursor" (on the enumerator) until they find an `<hr>`  (like the one for quotes or text boxes).
Having the handling of the enumerator distributed over several classes is probably not a good idea.

I can refactor the case statement into a registry-hash though...

PR #27, opened 2017-09-12
Add embedded element: Vimeo Video

dsager:
Embedded elements are specified by URL and can have optional tags following in brackets. The paragraph holds the caption.

Sorry for the big PR, but it's hard to split this up any further...


dsager:
Some problems with code climate, so I had to force push a few changes. I removed the duplicated comments from coveralls...

miguelbernadi:
+1

PR #28, opened 2017-09-12
Add embedded elements to parser

dsager:
### Support embedded elements in parser
Because the parser now detects embedded vimeo videos, the expected results of parsing the reference document in the integration tests had to be updated...

### Refactor embedded element class
Moving the to methods `#embed_id` and `.matches?` up into the base class, as they can be shared among all sub classes. This requires a rename of `.matches?` to `supported?`, which is less confusing anyway.


miguelbernadi:
+1

PR #31, opened 2017-09-12
Add embedded element: Tweets :bird:

dsager:


lib/article_json/import/google_doc/html/embedded_tweet_element.rb
```diff
@@ -0,0 +1,31 @@
...
+            match = @node.inner_text.strip.match(self.class.url_regexp)
+            "#{match[1]}/#{match[2]}" if match
+          end
+
+          class << self
+            # Regular expression to check if a given string is a Twitter URL
+            # Also used to extract the ID from the URL.
+            # @return [Regexp]
+            def url_regexp
+              %r(^\S*twitter\.com/?([^#/]+)(?:#|/status/|/statuses/)?(\d+))i
```
nicolas-fricke, line 24: You could also [name the matched groups here using the `?<group_name>` syntax](https://ruby-doc.org/core-1.9.3/Regexp.html#method-i-named_captures):
```ruby
%r(^\S*twitter\.com/?(?<handle>[^#/]+)(?:#|/status/|/statuses/)?(?<tweet_id>\d+))i
```
and then in `#embed_id` you could access them via name:
```ruby
"#{match[:handle]}/#{match[:tweet_id]}" if match
```

nicolas-fricke:
+1

ifosch:
+1 but I might consider @nicolas-fricke suggestion

lib/article_json/import/google_doc/html/embedded_tweet_element.rb
dsager: true, good idea for future refactoring...
although it would be necessarily easier to read :)

PR #32, opened 2017-09-12
Add embedded element: Facebook Video

dsager:


nicolas-fricke:
+1

EdLittle:
+1

ifosch:
+1 I guess @nicolas suggestion for twitter RegEx, would be valid here too

dsager:
@ifosch - in theory yes, but a little more overhead, as the id is generated in the parent class... I might refactor this all at once in the near future :)

PR #33, opened 2017-09-13
Add embedded element: Slideshare slide show

dsager:


bbrassart:
+1

EdLittle:
+1

ifosch:
+1

lib/article_json/import/google_doc/html/embedded_slideshare_element.rb
```diff
@@ -0,0 +1,35 @@
...
+          def embed_id
+            match = @node.inner_text.strip.match(self.class.url_regexp)
+            "#{match[:handle]}/#{match[:id]}" if match
+          end
+
+          class << self
+            # Regular expression to check if a given string is a Slideshare URL
+            # Also used to extract HANDLE and ID from the URL.
+            # @return [Regexp]
+            def url_regexp
```
towanda, line 23: just curious, why do you need this method to be a class method?

towanda:
+1

lib/article_json/import/google_doc/html/embedded_slideshare_element.rb
dsager: because another class method is calling it (in the base class):
https://github.com/Devex/article_json/blob/master/lib/article_json/import/google_doc/html/embedded_element.rb#L62

towanda: I see, thanks!

PR #34, opened 2017-09-13
Refactor embeddable element specs

dsager:
As the behavior of the specs is identical, we can move most of the logic into a shared context.


miguelbernadi:
+1

ifosch:
+1 Interesting concept this `shared_context`... I wonder if it is available for use with chefspec and if we could make any use of it.

dsager:
@ifosch this is a part of rspec-core, so I guess it should be available...
See https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-context

PR #36, opened 2017-09-13
Consistently use Nokogiri:HTML

dsager:
HTML nodes handle some things differently than XML nodes, for example href-links or mandatory closing tags...


nicolas-fricke:
+1

ifosch:
+1

PR #37, opened 2017-09-13
Separate element definition and parsing: Heading

dsager:
- Rename the `HeadingElement` to `HeadingParser`.


ifosch:
+1

miguelbernadi:
+1

PR #41, opened 2017-09-13
Separate element definition and parsing: List

dsager:
- Rename `ListElement` to `ListParser`


nicolas-fricke:
+1

bbrassart:
+1 I think

miguelbernadi:
+1

PR #42, opened 2017-09-13
Separate element definition and parsing: Image

dsager:
- Rename `ImageElement` to `ImageParser`
- Also removed the original image size, which was only used for debugging (therefore the parsed reference doc changed).


miguelbernadi:
+1

towanda:
+1

PR #43, opened 2017-09-13
Separate element definition and parsing: Text box

dsager:
- Rename `TextBoxElement` to `TextBoxParser`


towanda:
+1

PR #44, opened 2017-09-13
Separate element definition and parsing: Quote

dsager:
- Rename `QuoteElement` to `QuoteParser`


bbrassart:
+1

ifosch:
+1

PR #47, opened 2017-09-14
Extract embed element

dsager:
- Add `Embed` element, which can be re-used later on
- Additional test cases for `EmbeddedParser.build`
- Additional test cases for `EmbeddedParser.supported?`
- Rename some internal methods in `EmbeddedParser`


towanda:
+1

ifosch:
+1

PR #48, opened 2017-09-14
Code improvements

dsager:
- Extract `#float` and `#caption` methods: Several parsers support captions or floating, so the functionality can be shared among them via module...
- Extract the logic of collecting all nodes until a `<hr>` while advancing the enumerator into its own method.


towanda:
+1

ifosch:
+1

dsager:
rebased to resolve a conflict...

PR #49, opened 2017-09-14
Implement parsing of hashes

dsager:
This allows to initialize the intermediate format from hashes, which are obtained by parsing json code.

The added integration test parses the json fixture and "jsonifies" it to make result and source are identical.

This looks quite big, but it a lot of repetition and tests. Sorry :)


ifosch:
+1

miguelbernadi:
+1

PR #51, opened 2017-09-18
HTML export for text elements

dsager:
Text elements are exported to HTML, taking both links and styling into account.


nicolas-fricke:
+1

ifosch:
+1

PR #52, opened 2017-09-18
html export for paragraphs & lists

dsager:


towanda:
+1

ivandelrio:
+1

christofdamian:
I am not sure if I like creating html with nokogiri, this is pretty unreadable in my opinion.

dsager:
@christofdamian 
I prefer using a tool for the job instead of generating HTML code via string concatenation.
But you are right that it is harder to understand. I discussed this with @nicolas-fricke and we tried to design the code in a way that allows for easy changes later on (most of the stuff happens in a few central methods like `create_element`).

PR #54, opened 2017-09-18
html-export for text boxes, quotes and embeds

dsager:


dsager:
This looks a lot, but it's not that bad if reviewed commit by commit...

miguelbernadi:
+1

PR #57, opened 2017-09-19
Add custom configuration support and set oembed user agent

dsager:
#### Add top level configuration block
This allows the gem consumer to set configuration values the ruby-way:
```ruby
  config.oembed_user_agent = 'article-json oembed'
end
```

#### Set user agent header for oembed requests
Certain services (like facebook) require a user agent for oembed requests, now we can easily set one via config :tada:


miguelbernadi:
+1

ifosch:
+1

towanda:
+1

PR #58, opened 2017-09-19
Fix flaky configuration test

dsager:
Is the oembed test (which sets the user agent setting) runs first, the config test would fail. Now we always ensure an empty config.


towanda:
+1

miguelbernadi:
+1

ifosch:
+1

PR #61, opened 2017-09-20
Refactor element constructor pattern

dsager:
Slight change to the pattern to dynamically create exporter objects.

This is now aligned with the oembed resolver:
- `Base.build(element)` -> create a exporter object
- `Base.exporter_by_type(type)` -> find exporter class for element
- `Base#export` -> create a exporter object (if necessary) and call `#export` on it
- `*#export` -> export a nokogiri node


nicolas-fricke:
👍

towanda:
+1

PR #62, opened 2017-09-20
Add top level article class

dsager:
This class is meant as a single point of control for the different conversion from and to the article json format.


nicolas-fricke:
👍  looks good :)

towanda:
+1

dsager:
+1

PR #65, opened 2017-09-25
Add amp elements base and text

towanda:
As a first step into adding amp export support, we add element base and element
text.

### DIFF
--- a/lib/article_json/export/amp/elements/base.rb
+++ b/lib/article_json/export/amp/elements/base.rb
@@ -1,0 +1,52 @@
+module ArticleJSON
+  module Export
+    module AMP
+      module Elements
+        class Base
+          # @param [ArticleJSON::Elements::Base] element
+          def initialize(element)
+            @element = element
+          end
+
+          # Export a HTML node out of the given element
+          # Dynamically looks up the right export-element-class, instantiates it
+          # and then calls the #build method.
+          # @return [Nokogiri::HTML::Node]
+          def export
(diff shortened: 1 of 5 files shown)


lib/article_json/export/amp/elements/base.rb
```diff
@@ -0,0 +1,52 @@
...
+          # @return [Nokogiri::HTML::Node]
+          def export
+            exporter = self.class == Base ? self.class.build(@element) : self
+            exporter.export unless exporter.nil?
+          end
+
+          private
+
+          def create_element(tag, *args)
+            Nokogiri::HTML.fragment('').document.create_element(tag.to_s, *args)
```
dsager, line 23: I guess we can share this with the HTML exporter via module? same goes for `#create_text_node`...

towanda: yeah, that makes sense...

towanda: trying that, I'm closing this PR, since it's a different approach.

lib/article_json/export/amp/elements/text.rb
```diff
@@ -0,0 +1,44 @@
+module ArticleJSON
+  module Export
+    module AMP
+      module Elements
+        class Text < Base
```
dsager, line 5: Isn't this exactly the same as the HTML text element? Maybe we could just extend it?

PR #66, opened 2017-09-27
Readme small fix

towanda:


ivandelrio:
+1

miguelbernadi:
+1

dsager:
+1

usually, it's a good idea to add this to a `.rspec` file, then you don't need to specify it. not sure if we should commit this instead of the readme change?! @towanda

towanda:
@dsager I like much more the .rspec option, changing it

dsager:
oops, I merged it already, you'll have to do another PR :)

PR #69, opened 2017-09-27
Add amp elements base and text

towanda:
As a first step into adding amp export support, we add element base and element
text.


christofdamian:
+1

dsager:
@towanda 
there seems to be a let in your spec that's not used:
https://coveralls.io/builds/13454181/source?filename=spec%2Farticle_json%2Fexport%2Famp%2Felements%2Fbase_spec.rb

that's why the code coverage check fails...

towanda:
@dsager removing it, thanks :)

miguelbernadi:
+1

ifosch:
+1

dsager:
+1

PR #71, opened 2017-09-27
Add List and Heading AMP export element

towanda:


miguelbernadi:
+1

ifosch:
+1 but codeclimate is not happy @towanda

towanda:
@ifosch yes, but nothing we can do about it. We talked about this  morning in the standup, and first we are going to do duplicating some code, and later refactoring.

christofdamian:
+1 I think for this cases we can inherit from the HTML versions or point to them in the AMP exporter element configuration

ifosch:
ah, ok, @towanda, now I understand manual merges I've been seeing :tongue:

towanda:
@christofdamian good point, but at this point I feel more confident copying the code, and once that we have an implementation that's tested and then refactor.

dsager:
+1

PR #75, opened 2017-09-29
 Add AMP quote element 

towanda:
And also add make the Heading and List element to inherit from the HTML implementation.


dsager:
the spec for amp-paragraph still says `descibre ..HTML..Paragraph` :see_no_evil: 
that's why the code coverage dropped (before it was covered by the list spec)

dsager:
Due to problems with the namespace evaluation in sub-classes, we decided to keep the code duplicated. One benefit is that it will make sub-gems easier to pull of down the road. For now it will reduce our code-climate score :(

PR #77, opened 2017-10-02
Add AMP TextBox Element

towanda:


christofdamian:
+1 shall we kick codeclimate out?

dsager:
+1
I like to have the code climate, like this at least you have to double check if you are fine with the duplication...

towanda:
I agree with keeping CC, this way we know it.

miguelbernadi:
+1

nicolas-fricke:
+1 Looks good :)

PR #81, opened 2017-10-02
Change HTML & AMP export of quotes to <div>

dsager:
Since quotes are rendered within the article body, they shouldn't be an `aside` tag. Also, none of the other elements is using the specific HTML5 markup...


towanda:
+1, thanks!

christofdamian:
+1

PR #82, opened 2017-10-02
Add a default size for images

towanda:


lib/article_json/export/amp/elements/image.rb
```diff
@@ -6,6 +6,15 @@ class Image < Base
           include Shared::Caption
           include Shared::Float
 
+          # @param [ArticleJSON::Elements::Base] element
+          # @param String width
+          # @param String height
+          def initialize(element, width: '640px', height: '480px')
```
dsager: I think it should not have the `px` in the value, all the examples don't have it (as a normal `img` tag wouldn't):
https://www.ampproject.org/docs/reference/components/amp-img

nicolas-fricke: Also: where do these values come from and how can they be overwritten? 

If they shouldn't be the AMP-standard values, making them configurable through the gem config might be good..

towanda: @dsager right, changing it

towanda: @nicolas-fricke  we still don't know, the thing is that for AMP is mandatory to specify the image width and height, and we don't store it in the json structure yet. 

So the plan was to hardcore the default values we use in devex for the images, have one implementation working and then think about how to make it configurable. But I thought having an initializer could be nicer and more useful than having it just hardcored.

dsager:
@towanda you forgot to change the tests :)

towanda:
@dsager changed, I need a coffe :)

dsager:
@towanda 
Now that I think about it, I don't like the initializer options. Because it's the only element that overwrites the initializer, and the factory pattern we are using in the base class doesn't know about it. I would just add height/width methods that, for now, return the hard coded values. we can then later enhance it...
What do you think?

towanda:
@dsager I see your point, and I think it's ok for now, but at some point we'll need to implement a way to specify different sizes per image.

dsager:
Yes, but it should be part of the image element, not just the exporter. Like I said, I would just hard code it now - we can then later add size info to the format itself or have the default configurable. Important is, that before displaying, we'll scale the images accodingly using cloudinary.

towanda:
@dsager changed, thanks for the comments

dsager:
+1
Thanks!

yaneq:
+1

PR #84, opened 2017-10-04
Add facebook videos amp support

towanda:


christofdamian:
+1

yaneq:
+1

lib/article_json/export/amp/elements/embed.rb
```diff
@@ -20,20 +20,40 @@ def embed_node
             end
           end
 
-          def amp_tag
-            {
-              vimeo_video: 'amp-vimeo',
-              youtube_video: 'amp-youtube',
-            }[@element.embed_type.to_sym]
+          def type_specific_node
```
dsager, line 23:
```ruby
send("#{@element.embed_type}_node".to_sym)
```
:)

dsager:
+1

lib/article_json/export/amp/elements/embed.rb
towanda: mmm I prefer to declare the methods, the older I'm the less I like `send` :)

lib/article_json/export/amp/elements/embed.rb
```diff
@@ -20,20 +20,40 @@ def embed_node
...
+            when :facebook_video
+              facebook_node
+            end
           end
 
-          def type_specific_node
-            create_element(amp_tag,
+          def youtube_node
+            create_element('amp-youtube',
+                           'data-videoid' => @element.embed_id,
```
nicolas-fricke, line 36: Since the other keys are symbols here you can also use `'my-key': <value>` instead of `=>` and then `'data-videoid'` will also be a symbol-key :)

```ruby
create_element('amp-youtube',
               'data-videoid': @element.embed_id,
               width: default_width,
               height: default_height)
```

nicolas-fricke:
👍

lib/article_json/export/amp/elements/embed.rb
towanda: Oh, that's nice, I'll change it in next PR, thanks!

miguelbernadi:
+1

lib/article_json/export/amp/elements/embed.rb
towanda: @nicolas-fricke Changed here https://github.com/Devex/article_json/pull/85

towanda: I tried that, but code climate complains, because it doesn't seem compatible with ruby 2.1
https://codeclimate.com/github/Devex/article_json/pull/85

So I'm reverting it.

nicolas-fricke: Oh, I didn't know that the Gem was 2.1 compatible.. I thought that we'd only target >2.3 (I thought I've seen usages of the safe navigation operator `foo&.bar` somewhere, which is >2.3 only).

towanda: You are right, Daniel told me yesterday we are using ruby >= 2.3, but I got confused by the code climate error message.

PR #88, opened 2017-10-05
Fix `#nodes_until_hr`

dsager:
Check if there are more nodes before calling `peek`, which otherwise throws a `StopIteration` error.

Sometimes the document isn't perfectly formatted, in this case a block that is never closed by a horizontal line.

### DIFF
--- a/lib/article_json/import/google_doc/html/parser.rb
+++ b/lib/article_json/import/google_doc/html/parser.rb
@@ -105,8 +105,8 @@ module ArticleJSON
           # @return [Array[Nokogiri::HTML::Node]]
           def nodes_until_hr
             nodes = []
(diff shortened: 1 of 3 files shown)


miguelbernadi:
+1

christofdamian:
+1

christofdamian:
this probably warrants a test though

dsager:
@christofdamian see added commit...

christofdamian:
thanks. I still hate those fixtures.

dsager:
would you prefer to have the HTML directly in the test?

towanda:
+1

christofdamian:
Formatted html would help a bit. I guess it is easier to understand if you deal with it all the time.

PR #90, opened 2017-10-05
List all amp javascript files needed

towanda:


spec/article_json/export/amp/exporter_spec.rb
```diff
@@ -9,4 +9,29 @@
...
+        'src="https://cdn.ampproject.org/v0/amp-vimeo-0.1.js"></script>',
+        '<script async custom-element="amp-youtube" ' \
+        'src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>',
+        '<script async custom-element="amp-facebook" ' \
+        'src="https://cdn.ampproject.org/v0/amp-facebook-0.1.js"></script>',
+        '<script async custom-element="amp-iframe"' \
+        'src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>',
+        '<script async custom-element="amp-twitter" ' \
+        'src="https://cdn.ampproject.org/v0/amp-twitter-0.1.js"></script>'
+      ]
```
christofdamian, line 30: is the order in a Set defined?

lib/article_json/export/amp/exporter.rb
```diff
@@ -16,6 +17,16 @@ def html
           end
           doc.to_html(save_with: 0)
         end
+
+        def amp_libraries
```
dsager: can you please add some method documentation?! thanks!

towanda: sure!

lib/article_json/export/amp/elements/embed.rb
```diff
@@ -4,22 +4,16 @@ module AMP
       module Elements
         class Embed < Base
           include Shared::Caption
+          attr_reader :amp_library
```
dsager: what do you think about implementing this as a method:
```ruby
def amp_library
  return @amp_library if defined? @amp_library
  type_specific_node # the execution will determine needed amp library
  @amp_library
end
```

like this we would keep `#type_specific_node` private and the logic of when/how stuff is calculated withing this method...

spec/article_json/export/amp/exporter_spec.rb
nicolas-fricke: Nope, the order is uncertain:

> Set implements a collection of unordered values with no duplicates.

https://ruby-doc.org/stdlib-2.4.2/libdoc/set/rdoc/Set.html (very first line :wink: )

lib/article_json/export/amp/elements/embed.rb
towanda: I like it, changing it

spec/article_json/export/amp/exporter_spec.rb
christofdamian: then one should probably not compare these two arrays with ```eq``` below.

dsager: right, `match_array` is what we should use.

towanda: changed, thanks!

towanda:
@dsager I added the changes you suggested.

christofdamian:
+1

dsager:
+1
nice, thanks!

PR #93, opened 2017-10-10
Fix facebook embedded url

towanda:
The facebook video url was not absolute, and this caused the iframe not to load.


miguelbernadi:
+1

nicolas-fricke:
:+1: Good catch!

ivandelrio:
+1

bbrassart:
+1

dsager:
+1
would be good, if you can fix this as well for the youtube videos (it works now, but better safe than sorry). See https://github.com/dsager/article_json/commit/4708dad044a946546455d3e29c61f26a4361494e

towanda:
@dsager sure!

PR #95, opened 2017-10-11
Refactor AMP library resolving

nicolas-fricke:
AMP uses custom HTML tags for elements like iframes or embedded youtube videos. These elements each require a javascript to be loaded to be properly rendered by the browser. 

The new class `ArticleJSON::Export::AMP::CustomElementLibraryResolver` resolves the custom tags to a list of javascript libraries which then can be included on the page.

By adding `ArticleJSON::Export::AMP::Exporter#custom_element_tags` the user can now get a list of all used libraries (e.g. to check if one should manually include an additional `iframe` library for an element used outside of the article).

By refactoring the logic like this, we now also no longer need to parse the AMP element before being able to get a list of all its required custom elements.

The specs for `ArticleJSON::Export::AMP::Exporter.amp_libraries` still pass (only some spacing had to be fixed). Therefore, the (tested) API of the class remains the same.

`ArticleJSON::Export::AMP::Elements::Base.amp_library` (was untested) has been removed and replaced with `.custom_element_tags` which now returns a list of all custom tags used by the element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Export quotes as `<div>` instead of `<aside>`
 - Make Google Parser more fault tolerant
 - Support Vimeo videos with old flash player URLs
+- Export AMP
+- Return required libraries for AMP rendering
 
 ## 0.1.0 - 2017/09/20
 This is the very first release, with the following functionality:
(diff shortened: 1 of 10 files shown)


christofdamian:
we already discussed this with @towanda and @dsager when the first version was implemented. I prefer having the libraries close to where they are used and not in a central place.

PR #96, opened 2017-10-11
Refactor AMP library resolving

nicolas-fricke:
AMP uses custom HTML tags for elements like iframes or embedded youtube videos. These elements each require a javascript to be loaded to be properly rendered by the browser. 

The new class `ArticleJSON::Export::AMP::CustomElementLibraryResolver` resolves the custom tags to a list of javascript libraries which then can be included on the page.

By adding `ArticleJSON::Export::AMP::Exporter#custom_element_tags` the user can now get a list of all used libraries (e.g. to check if one should manually include an additional `iframe` library for an element used outside of the article).

By refactoring the logic like this, we now also no longer need to parse the AMP element before being able to get a list of all its required custom elements.

The specs for `ArticleJSON::Export::AMP::Exporter.amp_libraries` still pass (only some spacing had to be fixed). Therefore, the (tested) API of the class remains the same.

`ArticleJSON::Export::AMP::Elements::Base.amp_library` (was untested) has been removed and replaced with `.custom_element_tags` which now returns a list of all custom tags used by the element.

### DIFF
--- a/lib/article_json/export/amp/elements/embed.rb
+++ b/lib/article_json/export/amp/elements/embed.rb
@@ -12,20 +12,29 @@ module ArticleJSON
             end
           end
 
-          def amp_library
-            return @amp_library if defined? @amp_library
-            type_specific_node # the execution will determine needed amp library
-            @amp_library
+          # Custom element tags required for this embedded element
+          # @return [Array[Symbol]]
+          def custom_element_tags
+            case @element.embed_type.to_sym
+            when :youtube_video then %i(amp-youtube)
+            when :vimeo_video then %i(amp-vimeo)
+            when :facebook_video then %i(amp-facebook)
+            when :tweet then %i(amp-twitter)
+            when :slideshare then %i(amp-iframe)
+            else []
+            end
           end
 
           private
 
+          # @return [Nokogiri::XML::Element]
           def embed_node
             create_element(:div, class: 'embed').tap do |div|
               div.add_child(type_specific_node)
             end
           end
 
+          # @return [Nokogiri::XML::Element|nil]
           def type_specific_node
             case @element.embed_type.to_sym
             when :youtube_video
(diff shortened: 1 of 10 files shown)


lib/article_json/export/amp/elements/embed.rb
```diff
@@ -41,52 +49,28 @@ def type_specific_node
             end
           end
 
-          def youtube_library
-            '<script async custom-element="amp-youtube" ' \
-            'src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>'
-          end
-
           def youtube_node
```
dsager: a method comment with the return type would be nice :)

lib/article_json/export/amp/exporter.rb
```diff
@@ -5,29 +5,37 @@ class Exporter
...
-            doc.add_child(Elements::Base.new(element).export)
+          amp_elements.each do |exported_element|
+            doc.add_child(exported_element.export)
           end
           doc.to_html(save_with: 0)
         end
 
-        # Return an array with all the javascript libraries needed for some especial
-        # AMP tags (like amp-facebook or amp-iframe)
+        def custom_element_tags
```
dsager: missing method comment :)

dsager:
+1

lib/article_json/export/amp/elements/embed.rb
nicolas-fricke: Yea, I agree – though this is not really part of this PR 😉

christofdamian:
we already discussed this with @towanda and @dsager when the first version was implemented. I prefer having the libraries close to where they are used and not in a central place.

I am not sure why this was refactored at all.

nicolas-fricke:
@christofdamian 

After having tried to integrate the gem like this into our code, the current implementation caused some problems. The article content returned by the gem is normally integrated into a page which also contains other elements (e.g. a menu and in our case a paywall that uses an `amp-iframe`). With the current implementation, it hard to check if the gem already includes the iframe script (have to rely on regexes over the script tags). If it does, we wouldn't want to include the script again, but otherwise, we'll have to add it manually.

With the implementation proposed in this PR, we can 1) simply check the array of used custom tags and then 2) optionally use the `CustomElementLibraryResolver` to get additionally needed script tags.

Also, if we should get some other elements in the future which require a custom tag but are _not_ `Embed` elements (e.g. ads or custom lists) it makes it easier if resolving the custom tags to their respective scripts is in a central location (because e.g. the `amp-iframe` might be required by multiple elements).

I also checked in with @towanda and then paired with a bit with @dsager before going down this road and we saw this as the best option..

christofdamian:
But wouldn't it have been enough to just collect the library names instead also rewriting how they are collected in the AMP embed class? This seems to be 50% of the change.

nicolas-fricke:
Not really, as we also will need this for adding ads (we will add support for this to the gem next week). Also, this way we can use the same library resolving from outside of the gem (e.g. for getting the script tag for our paywall). 

And having to parse the whole element into a Nokogiri node (via `type_specific_node`) only to be able to see what library is needed and then storing the library in an instance variable as a side-effect of the node methods (e.g. `youtube_node`) seemed a bit confusing to me..

towanda:
+1

dsager:
When we initially discussed this I was in favor of having the element tags as the main reference for the libraries, instead of the script tags. Because it would be easier to use "from the outside", in case we need to check for the presence of elements. So I think this refactor is a good next step.

Given that @christofdamian didn't block or downvote this, I'll go ahead and merge?

PR #98, opened 2017-10-11
Support custom AMP exporters

nicolas-fricke:
For this, make defining of custom exporters in configuration generic:
By adding `ArticleJSON::Configuration#register_element_exporters_for` we can in the future allow the user to use the same method to also add custom exporters e.g. for AMP.

Adapt and deprecate the existing methods (`#register_html_element_exporter`, `#html_element_exporter`, and `#html_element_exporter=` which were defined via the `attr_accessor`) so that they still work as they currently do. These will be removed soon.

Similarly to custom HTML exporters, the user can now define custom AMP exporters by defining them as:
```ruby
  config.register_element_exporters_for(
    :amp,
    image: ArticleJSON::Export::AMP::Elements::ScaledImage
  )
end
```

Also use `ArticleJSON::Configuration#exporter_for` for HTML exporter lookup. Since the `#exporter_for` method always either returns a class or `nil`, there is no need for first checking if there is a custom exporter. Instead, `||` will return the default exporter if no custom exporter is returned.

### DIFF
--- a/lib/article_json/configuration.rb
+++ b/lib/article_json/configuration.rb
@@ -15,18 +15,65 @@ module ArticleJSON
 
   class Configuration
     attr_accessor :oembed_user_agent
-    attr_accessor :html_element_exporter
 
     def initialize
       @oembed_user_agent = nil
-      @html_element_exporter = {}
+      @custom_element_exporters = {}
     end
 
     # Register a new HTML element exporter or overwrite existing ones.
     # @param [Symbol] type
     # @param [Class] klass
+    # @deprecated Use `#register_element_exporters_for(:html, ...)` instead
     def register_html_element_exporter(type, klass)
-      @html_element_exporter[type.to_sym] = klass
+      register_element_exporters_for(:html, type => klass)
+    end
+
+    # Return custom HTML exporters
+    # @return [Hash[Symbol => Class]]
+    # @deprecated use `#exporter_for` instead
+    def html_element_exporter
+      @custom_element_exporters[:html] || {}
+    end
+
+    # Set custom HTML exporters
+    # @param [Hash[Symbol => Class]] value
+    # @deprecated use `#register_element_exporters_for(:html, ...)` instead
+    def html_element_exporter=(value)
+      @custom_element_exporters[:html] = value
+    end
+
+    # Register new element exporters or overwrite existing ones for a given
+    # exporter type.
+    # Usage example:
+    #  register_element_exporters_for(:html,
+    #                                 image: MyImageExporter,
+    #                                 advertisement: MyAdExporter)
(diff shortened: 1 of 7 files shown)


nicolas-fricke:
Hm, since we agreed on duplicating the code for the AMP exporter, `codeclimate` is now acting up.. @dsager is this the way we want to do it? Otherwise we'd need to export some exporter base functionality.. Otherwise, if we want to keep it like this, could you approve the violations? Thanks :)

christofdamian:
+1 can we use this to remove some of the duplication?

lib/article_json/configuration.rb
```diff
@@ -15,18 +15,65 @@ def configure
...
+    # @deprecated Use `#register_element_exporters_for(:html, ...)` instead
     def register_html_element_exporter(type, klass)
-      @html_element_exporter[type.to_sym] = klass
+      register_element_exporters_for(:html, type => klass)
+    end
+
+    # Return custom HTML exporters
+    # @return [Hash[Symbol => Class]]
+    # @deprecated use `#exporter_for` instead
+    def html_element_exporter
```
dsager, line 35: I think this should be called `exporters`, as it returns all of them...

lib/article_json/configuration.rb
```diff
@@ -15,18 +15,65 @@ def configure
...
+
+    # Register new element exporters or overwrite existing ones for a given
+    # exporter type.
+    # Usage example:
+    #  register_element_exporters_for(:html,
+    #                                 image: MyImageExporter,
+    #                                 advertisement: MyAdExporter)
+    # @param [Symbol] exporter
+    # @param [Hash[Symbol => Class]] type_class_mapping
+    def register_element_exporters_for(exporter, type_class_mapping)
```
dsager, line 54: I think I would like `#register_element_exporters` or `#register_element_exporters` better...

lib/article_json/configuration.rb
```diff
@@ -15,18 +15,65 @@ def configure
...
+
+      @custom_element_exporters[exporter.to_sym] ||= {}
+      @custom_element_exporters[exporter.to_sym].merge!(type_class_mapping)
+    end
+
+    # Get custom exporter class for a given exporter and element type
+    # @param [Symbol] exporter_type
+    # @param [Symbol] element_type
+    # @return [Class|nil]
+    def exporter_for(exporter_type, element_type)
```
dsager, line 75: maybe `#element_exporter_for` would be clearer, as we usually call the whole thing an exporter (HTML or AMP)?!

dsager:
+1
@nicolas-fricke - Looks good to me. Please consider the suggested naming-changes. But also no hard feelings if you merge it as is...

PR #99, opened 2017-10-13
Renames some configuration methods

towanda:
/cc @nicolas-fricke After having talked with @dsager about it, we thought these names are better. No hard opinions though, so we can discuss it and change it later.


dsager:
+1
not sure if we should rename the deprecates methods though?! Are they used anywhere outside of the gem?

ifosch:
+1

PR #100, opened 2017-10-16
Add support for placing additional elements into article

nicolas-fricke:
The new method `Article#place_additional_elements` allows a user to place additional elements into the article, like e.g. advertisements. These elements need to be defined within the user's own code.

Each element that is added this way will directly get placed in between paragraphs of the article. The method ensures that an additional element is never added before or after any node other than paragraphs (e.g. an image). The elements are added in the order you pass them into the method. If the article should not have enough spaces to place all the provided elements, they will be placed after the last element in the article.

Usage examples are provided in the updated `README.md`.


ifosch:
+1 but this is a pretty long PR

christofdamian:
+1

dsager:
+1

dsager:
@nicolas-fricke can you please also update the changelog?

towanda:
+1

PR #103, opened 2017-10-19
Fix NodeAnalyzer to accept quote and textbox definition as heading

nicolas-fricke:
Quotes or Text Boxes are to be defined in the Google Document as a paragraph prefixed with either the `Quote:` or the `Textbox:` word and suffixed by a horizontal line.

This commit allows the hotword to also be within a heading instead of insisting on it to be a regular paragraph.


towanda:
+1, Nice!

miguelbernadi:
+1

dsager:
+1

PR #104, opened 2017-10-19
Convert `<br>` tags in Google document to newline characters

nicolas-fricke:
If a text element from the Google document contains one (or multiple) `<br>` linebreak elements, substitute them as `"\n"` characters within the `Text` element.

For this, add `NodeAnalyzer#br?` analyzer to check if node is a `<br>` character or only contains whitespace characters and a line break.

Currently, the exporters will just ignore `"\n"` characters when rendering AMP or HTML. A following PR will add support for this to the exporters.

### DIFF
--- a/lib/article_json/import/google_doc/html/node_analyzer.rb
+++ b/lib/article_json/import/google_doc/html/node_analyzer.rb
@@ -23,7 +23,7 @@ module ArticleJSON
           # @return [Boolean]
           def empty?
             return @is_empty if defined? @is_empty
-            @is_empty = node.inner_text.strip.empty? && !image? && !hr?
+            @is_empty = node.inner_text.strip.empty? && !image? && !hr? && !br?
           end
 
           # Check if the node is a header tag between <h1> and <h5>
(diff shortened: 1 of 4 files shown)


miguelbernadi:
+1

christofdamian:
+1

lib/article_json/import/google_doc/html/node_analyzer.rb
```diff
@@ -105,6 +113,19 @@ def type
...
+
+          # Return true if the node only contains <br> nodes and empty text
+          # @return [Boolean]
+          def only_includes_brs?
+            return false unless node.inner_text.strip.empty?
+            tags = node.children.map(&:name)
+            # Check if it only contains <br> and text nodes
+            return false unless tags.all? { |tag| %w(br text).include? tag }
+            # Check if at least one is a `<br>` node
+            tags.any? { |tag| tag == 'br' }
```
dsager: I think `tags.include?('br')` would be nicer.

dsager:
+1

lib/article_json/import/google_doc/html/node_analyzer.rb
nicolas-fricke: Good point – I'll change it right away :)

nicolas-fricke: Changed :tada:

bbrassart:
+1

PR #105, opened 2017-10-20
Add support for linebreaks in HTML exporter

nicolas-fricke:
If the element content contains `"\n"` characters these will be transformed to `<br>` HTML tags.

For this, ensure that all `#export` return a `Nokogiri::XML::NodeSet` so that multiple nodes are supported as a return value. Therefore, change `Base#create_element` to wrap the generated element in a `NodeSet` and allow passing a block to the method to add additional fields onto the wrapped element (this avoids having to `tap` the element and then access the first element in the node set).

For more details, please refer to the individual commit messages.

RFC @dsager & @towanda

### DIFF
--- a/lib/article_json/export/html/elements/base.rb
+++ b/lib/article_json/export/html/elements/base.rb
@@ -11,7 +11,7 @@ module ArticleJSON
           # Export a HTML node out of the given element
           # Dynamically looks up the right export-element-class, instantiates it
           # and then calls the #build method.
-          # @return [Nokogiri::HTML::Node]
+          # @return [Nokogiri::XML::Element]
           def export
             exporter = self.class == Base ? self.class.build(@element) : self
             exporter.export unless exporter.nil?
@@ -19,8 +19,23 @@ module ArticleJSON
 
           private
 
+          # Create a Nokogiri Node with the given `tag`
+          # @param [Symbol] tag - type of the created element
+          # @param [*Object] args - additional arguments for the element
+          # @yield [Nokogiri::XML::Element] Optional block can be passed, will
+          #                                 be executed with the generated
+          #                                 element as a parameter. Return
+          #                                 value of this method will remain the
+          #                                 main element.
+          # @return [Nokogiri::XML::Element]
           def create_element(tag, *args)
-            Nokogiri::HTML.fragment('').document.create_element(tag.to_s, *args)
+            Nokogiri::HTML
+              .fragment('')
+              .document
+              .create_element(tag.to_s, *args)
+              .tap do |element|
+                yield(element) if block_given?
+              end
           end
 
           def create_text_node(text)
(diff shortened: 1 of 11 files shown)


lib/article_json/export/html/elements/base.rb
```diff
@@ -11,38 +11,46 @@ def initialize(element)
...
+
+          # @param [String] text
+          # @return [Nokogiri::XML::NodeSet]
           def create_text_node(text)
-            Nokogiri::HTML.fragment(text).children.first
+            Nokogiri::HTML.fragment(text).children
           end
 
-          class << self
+            class << self
```
dsager: this doesn't look right :)

lib/article_json/export/html/elements/base.rb
```diff
@@ -11,38 +11,46 @@ def initialize(element)
...
+          # @return [Nokogiri::XML::NodeSet]
+          def wrap_in_node_set(elements, document)
+            Nokogiri::XML::NodeSet.new(document).tap do |node_set|
+              elements.each { |element| node_set.push(element) }
+            end
+          end
+
+          # @param [String] text
+          # @return [Nokogiri::XML::NodeSet]
           def create_text_node(text)
```
dsager: the name isn't really descriptive any more... `create_text_nodes`?

lib/article_json/export/html/elements/base.rb
```diff
@@ -44,13 +44,7 @@ def wrap_in_node_set(elements, document)
...
           end
 
-          # @param [String] text
-          # @return [Nokogiri::XML::NodeSet]
-          def create_text_node(text)
-            Nokogiri::HTML.fragment(text).children
-          end
-
-            class << self
+          class << self
```
dsager: ok, ignore the previous comment :)

lib/article_json/export/html/elements/text.rb
```diff
@@ -36,8 +36,17 @@ def bold_and_italic_node
...
+            return create_text_nodes(@element.content) if @element.href.nil?
+            create_element(:a, href: @element.href) do |a|
+              a.add_child(create_text_nodes(@element.content))
+            end
+          end
+
+          # @param [Nokogiri::HTML::NodeSet] text
+          def create_text_nodes(text)
+            Nokogiri::HTML
+              .fragment(text.gsub(/\n/, '<br>'))
```
dsager, line 48: we don't want to add this to the super class's method, do we?

dsager:
Looks good to me.

dsager (approved):

lib/article_json/export/html/elements/text.rb
nicolas-fricke: I think it's good to only have it in here since only text nodes support line breaks at the moment

ifosch:
+1

PR #108, opened 2017-11-02
Extract common methods from AMP and HTML exporters

nicolas-fricke:
AMP is basically just a custom version of HTML. Therefore, the code of the two exporters is very similar. For better reusability (e.g. when implementing a Facebook Instant Article exporter), extract the code into a `Common` module which then is reused by both classes.

For reviewing, the code in the `Common` modules is extracted from the `HTML` exporters. It should be fairly straight forward to review this commit-by-commit as each commit extracts the common code into a module and reuses it.

The only changes that were made to the base code other than extracting the methods are:
- Use `.exporter_by_type` instead of referencing elements directly (this is also better since this way the elements can be overwritten via the configuration)
- Add a `.base_class?` check to see if an element is the instance of the base class since comparing the class is not working anymore once it is extracted to a module
- Add `Base#base_class` to get the base class of an element instance
- Add `.export_format` to check for custom element exporters so that the lookup method can be made reusable


nicolas-fricke:
[B]locked – This is a proposal for now and should be the basis of discussion. If we decide that extracting code like this is useful, I would do the same for all the other element classes.

lib/article_json/export/common/html/elements/base.rb
```diff
@@ -0,0 +1,85 @@
...
+                elements.each { |element| node_set.push(element) }
+              end
+            end
+
+            # Check if the current instance is an instance of the base class or
+            # of a specific element. Since this common module is in a different
+            # namespace, a simple `self.class == Base` check does not work.
+            # @return [Boolean]
+            def is_base_class?
+              self.class.name.end_with?('::Base')
```
towanda: Nicer than I thought!

towanda:
+1, I like it!

towanda (dismissed):

lib/article_json/export/common/html/elements/base.rb
```diff
@@ -0,0 +1,85 @@
...
+            end
+
+            module ClassMethods
+              # Instantiate the correct sub class for a given element
+              # @param [ArticleJSON::Elements::Base] element
+              # @return [ArticleJSON::Export::Common::HTML::Elements::Base]
```
dsager, line 68: Is this the right way to say "it returns an object implementing this module"?

lib/article_json/export/common/html/elements/base.rb
```diff
@@ -0,0 +1,85 @@
...
+              def build(element)
+                klass = exporter_by_type(element.type)
+                klass.new(element) unless klass.nil?
+              end
+
+              # Look up the correct exporter class based on the element type
+              # @param [Symbol] type
+              # @return [ArticleJSON::Export::Common::HTML::Elements::Base]
+              def exporter_by_type(type)
```
dsager, line 77: The line breaks in this method feel weird. what about this?
```ruby
def exporter_by_type(type)
  key = type.to_sym
  custom_exporter = 
    ArticleJSON.configuration.element_exporter_for(export_format, key)
  custom_exporter || default_exporter_mapping[key]
end
```

dsager (approved):
Nice, definitely looks better than the duplication.

lib/article_json/export/common/html/elements/base.rb
nicolas-fricke: I think so – any other ideas?

lib/article_json/export/common/html/elements/base.rb
nicolas-fricke: This also doesn't fit within the 80 character limit due to the 8 level deep indentation of this method. But found another way to split it up now – please check if you like it better now :)

nicolas-fricke:
@towanda @dsager Thanks for your positive feedback up to now :)

Based on this, I have now extracted the code of all AMP / HTML elements. Please have a look at it again and tell me if you approve of the changes. As mentioned in the description above, it's probably easiest to review this on a commit basis.

christofdamian:
+1 better than the duplication, though I am not a big fan of includes.

lib/article_json/export/common/html/elements/base.rb
dsager: yes, better! thanks!

lib/article_json/export/common/html/elements/shared/caption.rb
```diff
@@ -0,0 +1,31 @@
...
+                  @element.caption.each do |child_element|
+                    caption.add_child(text_class.new(child_element).export)
+                  end
+                end
+              end
+
+              private
+
+              # @return [ArticleJSON::Export::Common::HTML::Elements::Base]
+              def text_class
```
dsager: The name is a bit too generic for my taste, given that it's from a module. A class that uses this module could have a method `text_class` referring to CSS classes. So maybe something like `caption_text_exporter` or `caption_text_exporter_class`?

lib/article_json/export/common/html/shared/module_utils.rb
```diff
@@ -0,0 +1,24 @@
+module ArticleJSON
+  module Export
+    module Common
+      module HTML
+        module Shared
+          module ModuleUtils
```
dsager: I don't like having this in a separate module much. Took me quite a bit to understand what's going on and what the name `ModuleUtils` actually means :)

I would just leave it in the exporter module where it s used, much easier to grasp than a module loading another module that extends the base class :)

dsager (approved):
GJ! Just two comments on the code...

The one thing I don't like about this approach, now that I see it, is: The HTML export is basically empty and all its logic is in the common modules. Feels weird. But still better than the duplication and I can't think of a better way...

lib/article_json/export/common/html/shared/module_utils.rb
nicolas-fricke: The issue is that this is used in two modules: In the `Exporter` as well as in the `Elements::Base`. I first had the method duplicated in both modules, but that also didn't look right. Any better implementation ideas or naming suggestions?

lib/article_json/export/common/html/elements/shared/caption.rb
nicolas-fricke: Good point, I'll change this right away :)

lib/article_json/export/common/html/shared/module_utils.rb
dsager: Like discussed in person, better to define explicit `namespace` methods where needed...

lib/article_json/export/common/html/elements/shared/caption.rb
nicolas-fricke: Done :)

nicolas-fricke:
@dsager I tried to incorporate your ideas – please have another look at the PR 😃

lib/article_json/export/common/html/elements/base.rb
dsager: nope :)

towanda: for me sounds good

dsager (approved):
Looks good. A bit tricky to review because of the rebase. In the future it's maybe better to add separate commits and do a rebase/fixup right before merging the PR, especially if it's such a big one like this...

towanda:
+1, much better than duplication, GJ!

towanda (approved):

nicolas-fricke:
@dsager Good point, I'll work with `fixup` commits next time :)

Thanks for the reviews!

PR #110, opened 2017-11-07
Handle bad OEmbed responses and return message

nicolas-fricke:
When requesting an OEmbed object, handle the situation that the provider might return a non-successful response. In that case, render a message for the user explaining that there was an error with the embedded element.

For this, standardize the `OEbedResolver::Base#source_url` method and add a `#name` method which will be included in the text for the user.

The text for the user will be rendered within the `<figure>` element within a nested `span` with the class `unavailable-embed`. This also allows the consumer to have custom CSS for this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## WIP
+- Handle non-successful OEmbed responses
+
 ## 0.2.0 - 2017/11/03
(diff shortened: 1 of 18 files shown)


christofdamian:
+1

dsager (approved):

towanda (approved):

PR #111, opened 2017-11-08
Prepare 0.2.1 release

nicolas-fricke:
This release is only fixing issues with bad OEmbed responses which previously caused the entire exporting of the article to fail. Now it will render a message instead of the embedded element telling the viewer that the embedded element could not be loaded.


towanda:
+1

miguelbernadi:
+1

PR #114, opened 2017-11-09
Export Facebook text boxes surrounded by a horizontal line

nicolas-fricke:
Facebook does not support any text box element nor does it allow having a custom CSS for styling a given div / class (like our `text-box` class). Text boxes normally contain content that is not part of the main article, but has a supportive character. Therefore, it is important that the reader understands that the main content continues after this paragraph.

This commit inserts a horizontal line above and below the actual text box content which should provide this visual clue of "to-be-continued".

Floating content is not supported, therefore don't add any floating classes.


miguelbernadi:
+1

lib/article_json/export/facebook_instant_article/elements/text_box.rb
```diff
@@ -4,6 +4,35 @@ module FacebookInstantArticle
...
+              div.add_child(ascii_art_line_node)
+            end
+          end
+
+          private
+
+          # Returns a paragraph with the `ascii_art_line_text_element`. This
+          # gets inserted above and below the text box content
+          # @return [Nokogiri::XML::NodeSet]
+          def ascii_art_line_node
```
towanda, line 26: I love the name :)

towanda:
+1

PR #115, opened 2017-11-10
Improve support of empty captions

nicolas-fricke:
Sometimes, an author wants to include an image, a quote, tweet or other embedded object without having it succeeded with a caption. This is now supported by writing `[no-caption]` below the element. The generated JSON will contain `caption: []` on the element.

Therefore, also add exporter support for elements without caption for images, embeds, and quotes.  In this case, the exporter is now able to simply export the element without any caption tag.


christofdamian:
+1

dsager (approved):

PR #118, opened 2017-11-13
DVX-6275 Export Embedded elements for Facebook Instant Articles

towanda:
Export embedded elements (tweets, youtube videos, etc) for Instant Articles
following the reference document:
https://developers.facebook.com/docs/instant-articles/reference/embeds


miguelbernadi:
+1

christofdamian:
+1

lib/article_json/export/facebook_instant_article/elements/embed.rb
```diff
@@ -3,7 +3,40 @@ module Export
...
+            end
+          end
+
+          private
+
+          # Type specific object that should be embedded
+          # @return [Nokogiri::XML::Element]
+          def embed_node
+            if %i(facebook_video tweet).include? @element.embed_type.to_sym
+              default_node
```
dsager: I would expect the default to be in the `else` branch, and not for two specific cases...

lib/article_json/export/facebook_instant_article/elements/embed.rb
```diff
@@ -3,7 +3,40 @@ module Export
...
+            else
+              embedded_object
+            end
+          end
+
+          def default_node
```
dsager: what about naming this `iframe_node`?

dsager (approved):

lib/article_json/export/facebook_instant_article/elements/embed.rb
towanda: I changed the method name to 'iframe_node'.

lib/article_json/export/facebook_instant_article/elements/embed.rb
towanda: good point, changed

lib/article_json/export/facebook_instant_article/elements/embed.rb
dsager: thanks!

PR #119, opened 2017-11-14
Add a plain text exporter

nicolas-fricke:
As the name suggests, this exporter generates a plain text version of the article. Rich text elements like images, embeds or even text boxes are not being rendered.

For this, extract the reusable part of the `Base` concern and implement the element exporters for this format. There are no exporters for the skipped items. But via the custom exporter API it is easy to add them, if they should be required for some consumers.

### DIFF
--- a/lib/article_json/export/plain_text/elements/list.rb
+++ b/lib/article_json/export/plain_text/elements/list.rb
@@ -1,0 +1,50 @@
+module ArticleJSON
+  module Export
+    module PlainText
+      module Elements
+        class List < Base
+          # List of strings, one per line. If the list type is "ordered" then
+          # each line is preceded with a number, otherwise a dash. Newlines
+          # within a list item are indented accordingly.
+          # @return [String]
+          def export
+            "#{list}\n"
+          end
+
+          private
+
+          # Plain-text list of items using the right prefix
+          # @return [String]
+          def list
+            @element
+              .content
+              &.each_with_index
+              &.map do |content, index|
+                "#{prefix(index)} #{paragraph_exporter.new(content).export}"
+                  .gsub(/\n(?!$)/, newline_with_indention)
+              end&.join || ''
+          end
+
+          # Prefix used for every list item
+          # @param [String] index
+          def prefix(index)
+            @element.list_type == :ordered ? "#{index + 1}." : '-'
+          end
+
+          # Newline followed by the right amount of spaces to align with the
+          # indentation from the prefix
+          # @return [String]
+          def newline_with_indention
+            "\n " + ' ' * prefix(0).length
+          end
+
+          # Get the exporter class for paragraph elements
+          # @return [ArticleJSON::Export::Common::HTML::Elements::Base]
+          def paragraph_exporter
+            self.class.exporter_by_type(:paragraph)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/article_json/export/plain_text/elements/paragraph.rb
+++ b/lib/article_json/export/plain_text/elements/paragraph.rb
@@ -1,0 +1,33 @@
+module ArticleJSON
+  module Export
+    module PlainText
+      module Elements
+        class Paragraph < Base
+          # Plain text from the paragraph. Any formatting is disregarded.
+          # Followed by a newline.
+          # @return [String]
+          def export
+            "#{text}\n"
+          end
+
+          private
+
+          # Plain text of the paragraph
+          # @return [String]
+          def text
+            @element
+              .content
+              &.map { |text_element| text_exporter.new(text_element).export }
+              &.join
+          end
+
+          # Get the exporter class for text elements
+          # @return [ArticleJSON::Export::PlainText::Elements::Base]
+          def text_exporter
+            self.class.exporter_by_type(:text)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/article_json/export/plain_text/elements/quote.rb
+++ b/lib/article_json/export/plain_text/elements/quote.rb
@@ -1,0 +1,35 @@
+module ArticleJSON
+  module Export
+    module PlainText
+      module Elements
+        class Quote < Base
+          # Quotes are just rendered with a preceding blank line. If a caption
+          # is present, it is rendered below the quote indented with two dashes.
+          # @return [String]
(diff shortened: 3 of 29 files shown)


nicolas-fricke:
Please give some feedback about how you like [the plain text format](https://raw.githubusercontent.com/nicolas-fricke/article_json/0c93fa94668b181f451c8d1212da86e77a6cb35c/spec/fixtures/reference_document_exported.txt). 

I wasn't sure what to do with the `Image`, `Embed` and `TextBox` elements.. For now, the plain text would just ignore them. Another idea for `Image` and `Embed` would be to somehow render the caption, but this might be confusing if there is no actual item. We could also add the URL of the item, but that would not really be plain-text anymore (e.g. it wouldn't be usable for a screen-reader).

`TextBox` elements are also tricky as they break the flow of the text and if we were to export them using some ASCII-art (like we do for Facebook Instant Articles) it would again be weird for screen readers.

Leaving all three element types out renders – in my opinion – the best fluently readable plain text version of an article, and using custom exporters one can always add an exporter for text boxes.

What are your takes on this?

CC @towanda and @dsager

dsager:
Regarding the format:
Generally I think it's fine for a first version, we can always improve it later on. The question is what is the main purpose for this. If it's about screen reader compatibility and pure-text representation for processing (say classification or entity extraction), we should be fine w/o the embedded elements, images and text boxes.

We could prefixed quotes with a `>` like it is common in plain text emails, to distinguish them from normal paragraphs. What do you think?

Going to review the code in a bit :)

christofdamian:
I would include links to embedded items including images, otherwise some articles might be not useful.
Unless this format is mainly for indexing, then I would just include the caption.

nicolas-fricke:
Yes, this is the main question: What is this format meant for. In our case, I am currently adding it mainly for indexing / searching over it. Hence, links as well as image or embedded source URLs would be counter-productive. The same applies for text boxes, as in our case we use them most of the time to point to other articles. Therefore, this article's body would then match all the headlines of those related articles. Also, having looked at quite a few articles, captions of embedded elements are often misleading (they give credit to the photographers and rarely mention anything that is not mentioned in the text itself).

But as this is an open source project, it is hard to say what else this format _could_ be used for. I think a screen reader version or also using it as input for classification are reasonable.

And then the idea would be that for other use-cases (e.g. a human-readable plain text version – though I'm not sure what the use cases other than [Lynx](https://lynx.browser.org/) users would be :D ), one could simply add custom exporters for the relevant types and export e.g. images as links along with their caption.

towanda:
+1, Now we need this for indexing / searching, if we (or anyone) need anything else we can add it in the future.

towanda (approved):

christofdamian:
+1

lib/article_json/export/plain_text/elements/list.rb
```diff
@@ -0,0 +1,37 @@
...
+          end
+
+          private
+
+          def list
+            @element
+              .content
+              &.each_with_index
+              &.map do |content, index|
+                "#{prefix(index)} #{paragraph_exporter.new(content).export}"
```
dsager, line 23: We probably need to indent every new line withing the list item's content?
How would nested lists work? (Not sure if they are supported in the other exporters?)

lib/article_json/export/plain_text/elements/paragraph.rb
```diff
@@ -0,0 +1,28 @@
+module ArticleJSON
+  module Export
+    module PlainText
+      module Elements
+        class Paragraph < Base
+          # Plain text from the paragraph. Any formatting is disregarded.
+          # Followed by a newline.
+          # @return [String]
+          def export
+            text = @element
```
dsager: this should go into a separate method, to be consistent with the  pattern you use with `"#{list}\n"`...

lib/article_json/export/plain_text/elements/paragraph.rb
```diff
@@ -0,0 +1,28 @@
...
+              &.map { |text_element| text_exporter.new(text_element).export }
+              &.join
+            "#{text}\n"
+          end
+
+          private
+
+          # Get the exporter class for text elements
+          # @return [ArticleJSON::Export::PlainText::Elements::Base]
+          def text_exporter
```
dsager: not related to this PR, but I we can probably have a `method_missing` for `*_exporter` calls in the base class... :)

lib/article_json/export/plain_text/elements/quote.rb
```diff
@@ -0,0 +1,26 @@
+module ArticleJSON
+  module Export
+    module PlainText
+      module Elements
+        class Quote < Base
+          # Quotes are just rendered with a preceding blank line. If a caption
+          # is present, it is rendered below the quote indented with two dashes.
+          # @return [String]
+          def export
+            text = extract_text(@element.content)
```
dsager: this should go into a private method as well...

dsager (approved):

dsager:
We could also have one option to include/exclude links to embedded objects & images:
```ruby
  config.text_export_include_multimedia_urls = true
end
```
But as already said, this can be added whenever needed...

christofdamian:
Probably OK to have this simple one without formatting and then add a markdown exporter that is more readable and useful.

lib/article_json/export/plain_text/elements/paragraph.rb
nicolas-fricke: Haha, yea, that'd be possible.. or otherwise `base_class` also always works as it would then use the lookup based on the given element type :)

lib/article_json/export/plain_text/elements/list.rb
nicolas-fricke: I believe we currently do not support nested lists.. And I will add a fixup commit for the indented newlines :)

nicolas-fricke:
@dsager As suggested, I added two `fixup!` commits:
4797fb1 indents newlines in lists
b5fc872 extracts the parsing of paragraph and quotes into separate, private methods

dsager (approved):
nice, thanks for the changes...

nicolas-fricke:
Great, thanks for the reviews! Then I'm squashing the commits and merge the PR :)

PR #120, opened 2017-11-15
Adds embedded type class

towanda:
For the embedded elements, we add a class depending of the embed type.
(youtube_video, vimdeo_video, facebook_video, slideshare or tweet)


lib/article_json/export/common/html/elements/embed.rb
```diff
@@ -20,7 +20,8 @@ def export
             private
 
             def embed_node
-              create_element(:div, class: 'embed') do |div|
+              type = @element.embed_type || :type
```
nicolas-fricke: What is the `|| :type` statement meant to do? If there is no `embed_type` defined, it would be okay to not have a class for it, wouldn't it? (also, I think if there's no `embed_type` defined, the code would explode anyways at some other point..)

spec/article_json/export/amp/elements/embed_spec.rb
```diff
@@ -18,7 +18,7 @@
     context 'with a youtube video' do
       let(:source_element_embed_type) { :youtube_video }
       let(:expected_html) do
-        '<figure><div class="embed">' \
+        '<figure><div class="embed youtube_video">' \
```
nicolas-fricke: Not sure if it makes sense, but I at least wanted to talk about it: 

For CSS classes it's the de-facto community standard to use dashes instead of underscores – `.my-class` rather than `.my_class`. There are different reasons for this, e.g. explained [here](https://csswizardry.com/2010/12/css-camel-case-seriously-sucks/) and [here](https://stackoverflow.com/a/7561329/921530). 

So should we change our type classes to use dashes as well? Or is it nicer for consistency between the `article_json` JSON document and the classes to have them with an underscore here?

towanda: Oh, I wasn't sure about this, I asked @ivandelrio and he told me there was no problem in using underscore for css classes.
After reading those articles, I think it makes sense to keep the convention for each language (dashes for css and underscore for ruby), and it's quite easy to change this. 

In the ruby code using symbols with dashes seems very weird to me (`:youtube-video` for example).

lib/article_json/export/common/html/elements/embed.rb
towanda: ouch, this is a leftover from a debugging session, removing it.

spec/article_json/export/amp/elements/embed_spec.rb
nicolas-fricke: Yes, I like the idea sticking with the language specific best-practices, too 👍 

And changing it should be as easy as to add `.tr('_', '-')` to the `@element.embed_type` call, right?

towanda: Yes, fixing the specs takes a bit longer :)

towanda: @nicolas-fricke changed!

dsager (approved):

nicolas-fricke (approved):
Great – and thanks for the changes 👏

PR #121, opened 2017-11-15
Improve support of multiple calls to `Article#place_additional_elements`

nicolas-fricke:
When calling `ArticleJSON::Article#place_additional_elements` multiple times with additional elements, they were previously directly placed into the article. This led to additional elements being placed very close together, since the second iteration did not know that there was already an additional element inserted earlier.

Now, the class keeps track of what additional elements are to be placed into the article, but only when accessing the `Article#elements` they are actually injected into the article. If more additional elements are added, the structure will be recalculated to guarantee an even distribution of the additional elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - GDoc Importer: Support `[no-caption]` text, returns empty caption for element
 - Fix AMP export of Twitter tweets
 - Add a plain text exporter
+- Improve behavior of multiple calls to `Article#place_additional_elements`
 
 ## 0.2.1 - 2017/11/08
 **Fix**: Handle non-successful OEmbed responses by rendering message
--- a/lib/article_json/article.rb
+++ b/lib/article_json/article.rb
@@ -1,10 +1,26 @@
 module ArticleJSON
   class Article
-    attr_reader :elements
+    attr_reader :article_elements, :additional_elements
 
     # @param [Array[ArticleJSON::Elements::Base]] elements
     def initialize(elements)
-      @elements = elements
+      @article_elements = elements
+      @additional_elements = []
+    end
+
+    # All elements of this article with optional additional elements placed in
+    # between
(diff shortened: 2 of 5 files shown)


ifosch:
+1

christofdamian:
+1

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - GDoc Importer: Support `[no-caption]` text, returns empty caption for element
 - Fix AMP export of Twitter tweets
 - Add a plain text exporter
+- Support multiple calls to `Article#place_additional_elements`
```
dsager: It was already supported before, the results were just bad :)
So maybe

> Improve placement for multiple calls to `Article#place_additional_elements`

would be more descriptive?

lib/article_json/article.rb
```diff
@@ -60,13 +76,13 @@ def to_plain_text
...
-    # article.
+    # article. If the method is called multiple times, the order of additional
+    # elements is maintained.
     # @param [Object] additional_elements
     def place_additional_elements(additional_elements)
-      @elements =
-        ArticleJSON::Utils::AdditionalElementPlacer
-          .new(self, additional_elements)
-          .merge_elements
+      # Reset the `#elements` method memoization
```
dsager, line 83: neat!

dsager (approved):
Nice, GJ! :clap:

CHANGELOG.md
towanda: I agree..

towanda (approved):
Nice!

CHANGELOG.md
nicolas-fricke: Good point, changing it :)

nicolas-fricke:
☝️ Only updated the `CHANGELOG.md` (and commit message) [as suggested](https://github.com/Devex/article_json/pull/121#discussion_r151363524). Waiting for passing checks and then merge this. Thanks for all reviews!

PR #123, opened 2017-11-20
Rework algorithm to place additional elements

nicolas-fricke:
While the previous algorithm was better in saturating articles with additional elements it did not perform well when placing a single additional element into an article, this was always placed at the end.

Considering the known use-cases of this algorithm (placing ads or CTA forms within an article), it seems more common that single elements are placed than that the article is being saturated with additional elements.

For this, also add `#length` methods on `Text` and `Paragraph` elements: Similar to the `String#length` method these methods return the number of characters contained by the element. For compatibility with the `String` methods, also add the alias `#size`.

Outline of the new algorithm:
1. It calculates every how many characters (only considering paragraphs) an additional element should be inserted, given the number of additional elements provided
2. It then iterates over the existing elements within the article, inserting them into the resulting, merged article
3. As soon as it has passed the right amount characters, it checks if the current position is eligible for inserting an additional element (the previous element needs to be a paragraph and the following needs to be either a paragraph or a headline)
  a. If so, it inserts the additional element and recalculates in how many characters the next element should be inserted, based on the remaining number of characters and remaining additional elements
  b. If not, it keeps iterating over the article until it finds an eligible position and only then inserts the additional element and recalculates the number of characters until the next insertion
4. If there are still additional elements remaining which couldn't be inserted (due to not finding enough eligible positions), they are appended to the end of the article

### DIFF
--- a/lib/article_json/elements/paragraph.rb
+++ b/lib/article_json/elements/paragraph.rb
@@ -32,6 +32,16 @@ module ArticleJSON
         end
       end
 
+      # Return the sum of all characters within the content's text elements
+      # @return [Integer]
+      def length
+        return 0 if empty?
+        @content.reduce(0) do |sum, element|
+          sum + (element.respond_to?(:length) ? element.length : 0)
+        end
+      end
+      alias size length
+
       class << self
         # Create a paragraph element from Hash
         # @return [ArticleJSON::Elements::Paragraph]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add a plain text exporter
 - Improve behavior of multiple calls to `Article#place_additional_elements`
 - Remove deprecated `#register_html_element_exporter`, use `#register_element_exporters` instead
+- Rework algorithm to place additional elements to support better placement
 
 ## 0.2.1 - 2017/11/08
 **Fix**: Handle non-successful OEmbed responses by rendering message
--- a/lib/article_json/elements/text.rb
+++ b/lib/article_json/elements/text.rb
@@ -40,6 +40,14 @@ module ArticleJSON
         empty? || content.gsub(/[\s\u00A0]/, '').empty?
       end
 
+      # Get the number of characters contained by this text element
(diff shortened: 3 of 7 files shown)


towanda:
+1, Nice!

towanda (approved):

lib/article_json/elements/paragraph.rb
```diff
@@ -32,6 +32,16 @@ def blank?
         end
       end
 
+      # Return the sum of all characters within the content's text elements
+      # @return [Integer]
+      def length
+        return 0 if empty?
+        @content.reduce(0) do |sum, element|
+          sum + (element.respond_to?(:length) ? element.length : 0)
```
dsager, line 40: You could also write this as `sum + (element&.length || 0)`

dsager (approved):

lib/article_json/elements/paragraph.rb
nicolas-fricke: Sadly this won't work because the safe navigation operator `&.` works different than `ActiveSupport`'s `#try`: it only checks if the expression before evaluates to `nil`, if not it executes the provided method.

Example:
```ruby
require 'active_support'

# While both behave the same for `nil`:

nil.try(:foo)
# => nil

nil&.foo
# => nil

# And known methods:

class MyClass
  def my_method
    :hello_world
  end
end

my_instance = MyClass.new

my_instance.try(:my_method)
# => :hello_world

my_instance&.my_method
# => :hello_world

# The safe navigation operator raises a `NoMethodError` when calling an undefined method:

my_instance.try(:unknown_method)
# => nil

my_instance&.unknown_method
# => NoMethodError: undefined method `unknown_method' for  #<MyClass:0x007fd182884108>
```

nicolas-fricke:
Thanks for the reviews 🎉

lib/article_json/elements/paragraph.rb
dsager: True, I obviously didn't think that one through :D

PR #129, opened 2017-12-06
Fix issue when inserting elements into article with empty paragraph.

christofdamian:
When the article ended in an empty element, the code tried to add
more than addition elements than available, inserting a nil element.
This avoids the problem.

### DIFF
--- a/lib/article_json/utils/additional_element_placer.rb
+++ b/lib/article_json/utils/additional_element_placer.rb
@@ -47,6 +47,7 @@ module ArticleJSON
           .each_cons(2)
           .each_with_object([]) do |(element, next_element), result|
             result << element
+            next if remaining_elements.empty?
             characters_passed += element.length if element.respond_to?(:length)
             next_in -= element.length if element.respond_to?(:length)
             if insert_here?(next_in, element, next_element)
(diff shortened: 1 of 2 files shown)


towanda:
+1, good catch!

dsager (approved):

lib/article_json/utils/additional_element_placer.rb
```diff
@@ -47,6 +47,7 @@ def merge_elements
           .each_cons(2)
           .each_with_object([]) do |(element, next_element), result|
             result << element
+            next if remaining_elements.empty?
             characters_passed += element.length if element.respond_to?(:length)
             next_in -= element.length if element.respond_to?(:length)
```
christofdamian, line 52: I noticed that these two lines could also be:
```
  if element.respond_to?(:length)
    characters_passed += element.length
    next_in -= element.length
  end
```
to reduce complexity.

nicolas-fricke: Yup, that sounds like a good idea :)

nicolas-fricke (approved):
Very good catch! Thanks for the PR :)

PR #131, opened 2017-12-06
Prepare 0.3.2 release

nicolas-fricke:
Contains another fix to prevent `nil` elements when placing additional elements on articles that end with empty paragraphs added by @christofdamian in PR #129 – thanks for the fix!


dsager (approved):

ifosch:
+1

towanda (approved):

PR #132, opened 2018-04-11
Add support for embedding soundcloud oembed

christofdamian:
I have to confess that I haven't tested if this works yet. 
Adding another embed involves quite a few changes to files. 
This is for DVX-7041


spec/article_json/import/google_doc/html/embedded_soundcloud_parser_spec.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'embed_parser_shared'
+
+describe ArticleJSON::Import::GoogleDoc::HTML::EmbeddedSoundcloudParser do
+  include_context 'for an embed parser' do
+    let(:expected_embed_type) { :soundcloud }
+    let(:expected_embed_id) { 'rich-the-kid/plug-walk-1' }
+    let(:expected_tags) { %w(vimeo test) }
```
dsager: `s/vimeo/soundcloud/`?

christofdamian: good point,  or just test? what is it for anyway?

dsager: Acutally the values don't really matter: the shared example writes the tags into the HTML fixture and then expects them to be extracted properly. Might still be less confusing to not use `vimeo` here ;)

bbrassart:
+1 I think

spec/article_json/import/google_doc/html/embedded_soundcloud_parser_spec.rb
christofdamian: OK, I am going to do that. I guess I also have to add the support for AMP.

dsager: I'm not entirely sure, but I think AMP should work out of the box

christofdamian:
@dsager I fixed the tags and added an amp exporter

dsager (approved):
:+1:

christofdamian:
codeclimate is correct, but it isn't my fault!

dsager:
I approved the warnings :)

PR #133, opened 2018-04-12
Use default width for iframe if source is using a percentage

christofdamian:


ifosch:
+1 but coveralls is not very happy. Some adjustments to checks might be needed.

christofdamian:
Adding more coverage

bbrassart:
+1

ifosch:
still +1

dsager (approved):

PR #136, opened 2018-05-10
Only include slug from the soundcloud URL.

christofdamian:
Changed the regex to only except the slug part from the URL and
ignore any trailing characters.


towanda (approved):

nicolas-fricke (approved):
Ah, very good addition – thanks!

PR #138, opened 2018-10-30
Import image href from caption text using a custom tag

towanda:
Although google docs lets the user to create images with an href, it doesn't
export them to html.
(See https://productforums.google.com/forum/#!topic/docs/85jCFF6McsU)

As a workaround for this, we create a custom tag [image-link-to: url] that when
used in the next paragraph of an image (which we actually use as the caption of
the image) we extract the url and pass it to the Image Element.

Next step would be to export this to the different formats.


lib/article_json/elements/image.rb
```diff
@@ -1,16 +1,17 @@
...
   module Elements
     class Image < Base
-      attr_reader :source_url, :caption, :float
+      attr_reader :source_url, :caption, :float, :href
 
       # @param [String] source_url
       # @param [Array[ArticleJSON::Elements::Text]] caption
       # @param [Symbol] float
-      def initialize(source_url:, caption:, float: nil)
+      def initialize(source_url:, caption:, float: nil, href: nil)
```
nicolas-fricke, line 10: Could you please also add the YARDoc `@param` annotation for `href`? Thanks :)

lib/article_json/import/google_doc/html/image_parser.rb
```diff
@@ -36,17 +37,42 @@ def float
...
           end
 
           private
 
+          # Removes the [image-link-to: url] tag from the caption node
+          def remove_image_link_tag
+            @caption_node
+              .children
+              .first
+              .content = @caption_node.content.gsub(href_regexp, '').strip
```
nicolas-fricke: As we only are able to ever parse one `[image-link-to: url]` per image caption, we could consider using `String#sub` instead of `String#gsub`. This could be slightly more performant and would keep repeated `[image-link-to: url]` mentions in the caption, which would maybe make it clearer to the user that something isn't going as expected. What do you think?

nicolas-fricke (changes requested):
Looks good so far! I just have a few small remarks.

Also, it might be good to add an integration test in [`spec/article_json/import/google_doc/html/parser_spec.rb` (similar to the `when paragraphs contain no text` context)](https://github.com/Devex/article_json/blob/8239230e38ba52d7e1ef14ba6dd22790e03d45c8/spec/article_json/import/google_doc/html/parser_spec.rb#L53-L77) with an actual Google Document fixture. That way, we could ensure that the whole process works and that the caption no longer contains the `image-link-to` label.

lib/article_json/elements/image.rb
towanda: Yes, I forgot about it

lib/article_json/import/google_doc/html/image_parser.rb
towanda: I really like it!

lib/article_json/import/google_doc/html/image_parser.rb
```diff
@@ -36,17 +37,42 @@ def float
...
+          end
+
+          # Regular expression to chech if there's a [image-link-to: url] tag
```
christofdamian: s/chech/check/

towanda: Ouch! good catch!

lib/article_json/import/google_doc/html/image_parser.rb
```diff
@@ -36,17 +37,42 @@ def float
...
+          # @return [Regexp]
+          def href_regexp
+            %r{\[image-link-to: (?<url>.*)\]}
```
christofdamian: I usually use `%r{\[image-link-to:\s+(?<url>.*)\]}` to make spaces a bit more flexible.

christofdamian:
+1

lib/article_json/import/google_doc/html/image_parser.rb
towanda: cool, changing it!

nicolas-fricke: Oh, just realized that since regexes are by default greedy, right now, `(?<url>.*)` would include everything in the `url` match up until the _last_ closing `]` in the string. Therefore, a caption like:
```ruby
string = '[image-link-to: https://www.devex.com] [no-caption]'
regex = %r{\[image-link-to: (?<url>.*)\]}

# This would result in:
regex.match(string)
# => #<MatchData "[image-link-to: https://www.devex.com] [no-caption]" url:"https://www.devex.com] [no-caption">
```
So our extracted URL would be `https://www.devex.com] [no-caption`.

What we can do instead is change the regex to either use the character class `[^\]]*` to include everything but the closing bracket:
```ruby
%r{\[image-link-to:\s+(?<url>[^\]]*)\]}
```
Or make the <url> part non-greedy using `.*?` instead of `.*`:
```ruby
%r{\[image-link-to:\s+(?<url>*?)\]}
```

It would be good to have this also covered by a unit test where our label contains another closing bracket and we expect the link to _not_ contain it.

nicolas-fricke (changes requested):

lib/article_json/import/google_doc/html/image_parser.rb
towanda: Oh, good catch! changin it.

lib/article_json/elements/image.rb
nicolas-fricke: Thanks for adding it :) Ideally, we should also add a bit more description to all those params, but that can be another PR that improves the `@param` annotations all around the project.

nicolas-fricke (approved):
Great, thanks for the changes!

PR #139, opened 2018-11-05
Export the image element href attribute as a link

towanda:
If the image element has an href attribute, we nest the figure tag of the image
inside a link.

### DIFF
--- a/lib/article_json/export/common/html/elements/image.rb
+++ b/lib/article_json/export/common/html/elements/image.rb
@@ -7,22 +7,36 @@ module ArticleJSON
             include ArticleJSON::Export::Common::HTML::Elements::Shared::Caption
             include ArticleJSON::Export::Common::HTML::Elements::Shared::Float
 
-            # Generate the `<figure>` node containing the image and caption
+            # Generate the `<figure>` node containing the image and caption or
+            # an `<a>` node containing the `<figure>` node if href is present.
             # @return [Nokogiri::XML::Element]
             def export
+              figure_node
+            end
+
+            private
+
+            # @return [Nokogiri::XML::NodeSet]
+            def figure_node
               create_element(:figure, node_opts) do |figure|
-                figure.add_child(image_node)
+                node =  @element.href.present? ? href_node : image_node
+                figure.add_child(node)
                 if @element.caption&.any?
                   figure.add_child(caption_node(:figcaption))
                 end
               end
             end
 
-            private
-
             # @return [Nokogiri::XML::NodeSet]
             def image_node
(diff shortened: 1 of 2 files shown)


lib/article_json/export/common/html/elements/image.rb
```diff
@@ -18,13 +26,18 @@ def export
               end
             end
 
-            private
```
nicolas-fricke, line 21: Was there any reason to make all methods public? If not, I like the idea of keeping the public API of our classes as small as possible and therefore keeping all these methods `private`.

lib/article_json/export/common/html/elements/image.rb
```diff
@@ -18,13 +26,18 @@ def export
...
-
             # @return [Nokogiri::XML::NodeSet]
             def image_node
               create_element(:img, src: @element.source_url)
             end
 
+            # @return [Nokogiri::XML::NodeSet]
+            def href_node
+              create_element(:a, href: @element.href) do |a|
+                a.add_child(figure_node)
```
nicolas-fricke: What would currently happen if the figure had a caption? Would the caption also be wrapped in the `<a>` tag? I am mainly asking because there may be other links in the caption, which would then mean nested `<a>` tags (and I have no clue how different browsers would handle that..). Therefore, I'd recommend only nesting the `` tag within the `<a>` and keeping the caption outside.

It'd also be good to have a test for this to see what happens if an image has both an `href` and a `caption`.

towanda: Good point!

lib/article_json/export/common/html/elements/image.rb
towanda: No, this was a mistake, changing it...

towanda: Sorry I got confused by your comment, but the private keyword is not removed, it's added before:
https://github.com/Devex/article_json/pull/139/files#diff-5f912eaf379ce4c0a5c00289a1442812R17

nicolas-fricke: Ouch, right – I must have been blind 🙈 Sorry!

lib/article_json/export/common/html/elements/image.rb
towanda: @nicolas-fricke Changed!

nicolas-fricke: Great, thanks!

nicolas-fricke (approved):
Looks good, thanks for the changes! :)

PR #140, opened 2018-12-12
Prepare release 0.3.5

towanda:
**Improvements** to import and export image links from Google Docs
- Import image `href`` from caption text using a custom tag
- Export the image element href attribute as a link


nicolas-fricke (approved):
Great, thanks for the changes! :)

PR #141, opened 2019-08-02
Fix and update fixtures

towanda:
As the api for embeding tweets, vimeo and youtube videos has change,
we need to update the fixtures used for mocking requests and also
updating the reference documents for testing.

Also fixes a bug caused by using .present? without active support.


dsager (approved):
LGTM

christofdamian:
If GitHub doesn't show a diff, then the fixtures are probably unreadable. 
I guess they are fine for integration tests, but now I don't know if they were updated in any meaningful way for the tests to pass.

towanda:
@christofdamian The fixtures are readable, and gh shows the diff. The ones which are not the are the reference documents, which are generated by the [update_reference_script.sh](https://github.com/Devex/article_json/blob/master/bin/update_reference_document.sh)

Maybe we could update the testing process so that they are generated each time? but that could make the test suite slower.

Anyway, this has been like this since the beginning of the project, this time we needed to change because as the responses from some embedded apis have changes, the whole thing changes. I'm really open to change this, but might be a bigger effort, and already took me sometime to realize what was happening (I made a small change, and some test which seemed not related failed).

christofdamian:
@towanda I realize it was like this before. It wasn't meant as critic on the current PR. Running the update reference script in CI might be an idea.

nicolas-fricke (approved):
Looks good, thanks for the update

PR #142, opened 2019-08-05
Add tag support for text box element

towanda:
We add support for tags in the text box element, in the same way as we
did for the embedded elements.


christofdamian:
+1

miguelbernadi:
+1

lib/article_json/import/google_doc/html/text_box_parser.rb
```diff
@@ -22,10 +22,21 @@ def content
             @nodes.map { |node| parse_sub_node(node) }.compact
           end
 
+          # Extract any potential tags, specified in brackets after the URL
```
dsager: what URL? :thinking: 
```suggestion
          # Extract any potential tags, specified in brackets after the Textbox definition
```

towanda: You are right, changed!

dsager: From the test it looks like they come at the beginning of the box content?

dsager (approved):

lib/article_json/import/google_doc/html/text_box_parser.rb
```diff
@@ -22,10 +22,21 @@ def content
             @nodes.map { |node| parse_sub_node(node) }.compact
           end
 
+          # Extract any potential tags, specified in brackets after the Textbox definition
+          # @return [Array[Symbol]]
+          def tags
+            match = /(.*?)[\s\u00A0]+\[(?<tags>.*)\]/.match(@float_node.inner_text)
```
nicolas-fricke, line 28: Would this only match words in brackets in the first row (expected behavior) or would it match any brackets in a textbox DOM? We'd only expect the tags to be picked up in example 1, but not in 2, right?

```
Textbox: [tag-1 tag-2]
Some textbox content...
```
vs
```
Textbox:
Some textbox content... [tag-1 tag-2]
```

lib/article_json/import/google_doc/html/text_box_parser.rb
```diff
@@ -22,10 +22,21 @@ def content
...
+            (match ? match[:tags] : '').split(' ')
```
nicolas-fricke: Not a blocker, but I'd prefer a guard clause to make the logic a bit easier to grasp:
```suggestion
            return [] unless match
            match[:tags].split(' ')
```

nicolas-fricke (commented):
It would be great to have some tags on one of the two textboxes in the reference document (or have a third textbox that only shows this feature off) and then also have it be part of the integration test of the reference document :)

lib/article_json/import/google_doc/html/text_box_parser.rb
towanda: Yes, you are right, it would work like in example 1.

towanda:
@nicolas-fricke Good point, on it.

towanda:
@dsager @nicolas-fricke @miguelbernadi @christofdamian As I had to make more changes due to a bug, I'm creating a new [PR](https://github.com/Devex/article_json/pull/143) and closing this.

PR #143, opened 2019-08-05
 Add tag support for text box element

towanda:
We add support for tags in the text box element, in the same way as we
did for the embedded elements.


ifosch:
+1

dsager:
The last commit message is a bit generic, regarding the changes that are happening. Would be good to explain why we need to pass in the current node, for example...

But I guess que can just autosquash this when merging...

dsager (approved):

lib/article_json/import/google_doc/html/text_box_parser.rb
```diff
@@ -5,14 +5,16 @@ module HTML
         class TextBoxParser
           include Shared::Float
 
+          # @param [Nokogiri::HTML::Node] Current node
```
nicolas-fricke: It took a bit to understand why we pass a `current_node` in and what the difference of this and `nodes` is. Some explanation could help :) 

Also, I think as per our version of YARD what comes after `@param [...]` needs to be the parameter name. So this could be something like:
```suggestion
          # @param [Nokogiri::HTML::Node] current_node Document node that states
                                                       that this is a textbox. 
                                                       May contain tags, too.
```

nicolas-fricke: Another idea might be to even rename the param to something like `type_node` or `first_node`...

towanda:
@nicolas-fricke I made the changes you suggested.

nicolas-fricke (approved):
Great, thanks for the changes 👍

dsager:
Thanks for the changes and additions!

PR #145, opened 2019-08-20
Use https for soundcloud oembed api

towanda:
After the last change to soundclued oembed, a plain http call would
send a redirection to an https request, making the player to fail. To
fix this, we use https instead.


christofdamian:
+1

dsager (approved):

nicolas-fricke (approved):
Thanks for fixing this!

PR #147, opened 2020-03-23
NUT-113 Update fixtures and stubbed oembed request

towanda:
Updates the fixtures for reference exported documents, stubbed oembed requests and also adds an script to update all the reference exported documents.


christofdamian:
+1

christofdamian (approved):

dsager (approved):

dsager:
LGTM, thanks!
I think we should probably also add something to the README, explaining when and how to run the scripts?

PR #148, opened 2020-03-24
Add alt attribute to images

towanda:
Adds an "alt" attribute to the image element, and also adds it to the google doc
importer and to the different exporters we support.


christofdamian (approved):
+1 I didn't read the the reference doc diff :-)

dsager:
+1

dsager (approved):

PR #149, opened 2020-04-02
Update oembed fixtures and reference documents

towanda:
Updates the fixtures used to mock the response from some oembed requests and the
reference exported documents.

Also updates the README file with information about how to update the reference
documents and the stubbed oembed requests for tests.


miguelbernadi:
+1

dsager (approved):

PR #150, opened 2020-04-06
Fix image link to tag

towanda:
Fixes the use of the tag [image-link-to: url] to generate a link for an
image. The problem was that when Google exports a document to html, if the
paragraph has an image link tag, it creates three different spans. For an
original paragraph like this:

`This is an example of [image-link-to: http://devex.com] tag`

Google will export it as:

`<span>This is an example of [image-link-to:</span>`
`<span><a href=http://devex.com>http://devex.com</a></span>`
`<span>] tag</span>`

So we need to remove the image link tag from the three spans, and also to check
for the [no-caption] tag in a different way.


ifosch:
@towanda Apparently, there's a chained spec failing. Not sure if that requires just some more fixture updating or decoupling these specs, if possible.

christofdamian:
+1 looks like the google id changed, this one is expected now: `DjKv5wkG-CjXExV-RnBJwqicDQr3...`

towanda:
@ifosch @christofdamian I changed the document I was using for testing image links, but forgot to change the test. Now it's updated.

dsager (approved):

christofdamian:
still +1

PR #151, opened 2020-04-30
Fix [image-link-to: ] tag removal

towanda:
When trying to remove the [image-link-to: ] tag after parsing an image caption,
we used the last element of the caption childrens, so if there was more text
after the tag, we tried to remove the `]` from the text at the end instead of
from the proper child.

### DIFF
--- a/spec/article_json/import/google_doc/html/image_parser_spec.rb
+++ b/spec/article_json/import/google_doc/html/image_parser_spec.rb
@@ -184,17 +184,19 @@ describe ArticleJSON::Import::GoogleDoc::HTML::ImageParser do
 
     context 'when the caption is `[image-link-to]`' do
       let(:text) { 'Original caption text' }
+      let(:second_text) { 'second text' }
       let(:caption_fragment) do
         "<p><span> [image-link-to: </span>" \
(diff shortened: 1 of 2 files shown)


spec/article_json/import/google_doc/html/image_parser_spec.rb
```diff
@@ -184,17 +184,19 @@
 
     context 'when the caption is `[image-link-to]`' do
       let(:text) { 'Original caption text' }
+      let(:second_text) { 'second text' }
       let(:caption_fragment) do
         "<p><span> [image-link-to: </span>" \
         "<span><a href=\"http://devex.com\">http://devex.com</a></span>" \
-        "<span>] #{text}</span></p>"
+        "<span>] #{text}</span><span>#{second_text}</p>"
```
bbrassart: It probably doesn't make a difference for the test but the second `span` should probably be closed, right?

towanda: Oh, good point, fixing it. thanks!

towanda: Changed!

bbrassart:
+1 I think

ifosch:
+1

PR #152, opened 2020-05-13
Revert last changes of image-link-to tag

towanda:
The last changes about the image-link-to tag tried to fix the changes than google doc does to some elements, like converting a plain url into an href html tag surrounded by spans. 

Trying to adapt to these google doc changes is too complicated, so we are trying to use plain text in the caption, forcing editors to paste the caption text without formatting.


towanda:
Fixing tests.

christofdamian:
+1

miguelbernadi:
+1

PR #155, opened 2020-11-19
Add support for the new facebook oembed endpoints

towanda:
We add the option to access the new facebook oembed endpoints, which are not
public anymore. To achieve this, we add an option in the configuration class to
add the token we need to access this endpoints.

Also we remove the facebook video from the reference document, so that we can run all the tests without having to use a facebook token.


christofdamian:
+1

spec/article_json/utils/o_embed_resolver/facebook_video_spec.rb
```diff
@@ -1,21 +1,49 @@
...
+      let(:expected_name) { 'Facebook video' }
     end
-    let(:oembed_response) do
-      File.read('spec/fixtures/facebook_video_oembed.json')
+  end
+
+  describe '#access_token' do
+    subject { described_class.build(element).access_token }
+
+    context 'when the config has not the token' do
```
dsager, line 33:
```suggestion
    context 'when the config has no token' do
```

dsager:
+1

miguelbernadi:
+1

ifosch:
+1

PR #156, opened 2020-11-23
Fixes google doc import format

towanda:
Google docs changed again the export to html format. Now we check the presence
of a div element, to parse the document both ways (with all content nested under
a div, or directly under the body tag).


miguelbernadi:
+1

dsager:
+1

rraymundo:
+1

christofdamian:
+1 they changed back to the previous format? Interesting.

PR #158, opened 2020-11-25
Add an script to check the google doc export format

towanda:
Adds an script that downloads an exported version of the reference document and
checks that we can parse it correctly.

This way we can easily know if google has made any change to the export format.


miguelbernadi:
+1 Shouldn't we set some periodic execution to detect this as well? (I'd go with once a day for starters).

dsager:
+1

PR #159, opened 2020-11-25
Add github action to run the check script

towanda:
Adds a github action that runs the script to check the google export format hasn't changed.

### DIFF
--- a/ruby.yml
+++ b/ruby.yml
@@ -1,0 +1,29 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
+# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
+
+name: Ruby
+on:
+  schedule:
+  - cron: "0 6 * * *"
+  workflow_dispatch:
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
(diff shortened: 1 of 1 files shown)


ruby.yml
```diff
@@ -0,0 +1,29 @@
...
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
+# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
+
+name: Ruby
+on:
+  schedule:
+  - cron: "0 6 * * *"
+  workflow_dispatch:
```
miguelbernadi, line 12: What does this `workflow_dispatch` option do? Looks like a leftover.

miguelbernadi:
+1 Nice!

ruby.yml
towanda: This enables the  option to run the workflow manually, from [the documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_dispatch):

```
Manually running a workflow

To manually run a workflow, you can configure your workflow to use the workflow_dispatch event. This enables a "Run workflow" button on the Actions tab.

on:
  workflow_dispatch:
```

ruby.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
```
miguelbernadi: Shouldn't the `uses` begin with a dash?
```suggestion
      - uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
```

towanda: Good catch!

dsager:
+1 neat!

christofdamian:
+1

PR #171, opened 2022-05-17
BAN-1718 New param with to allow custom placer

xavi666:
With this optional param we can pass a different placer that has a different algorithm to distribute the elements in the article.

When not passing anything it will use the default one: `ArticleJSON::Utils::AdditionalElementPlacer` so this change won't break anything.

[BAN-1718 Edit article_json gem to make element location configurable](https://mydevex.atlassian.net/browse/BAN-1718)

### DIFF
--- a/lib/article_json/article.rb
+++ b/lib/article_json/article.rb
@@ -14,7 +14,7 @@ module ArticleJSON
     def elements
       @elements ||= begin
         if @additional_elements.any?
-          ArticleJSON::Utils::AdditionalElementPlacer
+          @additional_element_placer_class
             .new(@article_elements, @additional_elements)
             .merge_elements
         else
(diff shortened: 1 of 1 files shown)


xavi666:
Paired with @dsager 
FYI @damiendillon @EdLittle

lib/article_json/article.rb
```diff
@@ -91,10 +91,15 @@ def to_plain_text
     # article. If the method is called multiple times, the order of additional
     # elements is maintained.
     # @param [Object] additional_elements
-    def place_additional_elements(additional_elements)
+    # @param [ArticleJSON::Utils] with
```
xavi666: I am unsure how to define here the param type :thinking: @Devex/back-end-guild

dsager: I think you need to define a duck type here (walks like a duck, quacks like a duck, it probably is a duck!).
So, to say that you expect a class that responds to `#merge_elements`, you'd do this:
```suggestion
    # @param [Class<#merge_elements>] with
```
This still doesn't cover the expected signature for `#initialize`, but you can specify this in the free-text comment, that should be good enough...

xavi666: let's see if that looks better

lib/article_json/article.rb
```diff
@@ -91,10 +91,15 @@ def to_plain_text
     # article. If the method is called multiple times, the order of additional
     # elements is maintained.
     # @param [Object] additional_elements
-    def place_additional_elements(additional_elements)
+    # @param [Class<#merge_elements>] with - #initialize should be defined in the passed class
```
dsager:
```suggestion
    # @param [Class<#merge_elements>] with - The passes class's `#initialize` method needs
    #                                        to accept two lists of elements. See
    #                                        `ArticleJSON::Utils::AdditionalElementPlacer`
    #                                        for reference.
```

dsager:
+1

PR #173, opened 2023-04-21
Add support for adding images via URL

dsager:
Recently Google changed how images of an HTML export are accessed, the URLs became much longer and in certain cases require an active session to be accessed.

To allow for an alternative way of embedding images in news articles, we now added support to specify external images via the URL:
- When the image's URL is left aligned, it will be displayed as a small image with text floating around it.
- By aligning image URLs to the right they will later on be displayed on the right side of the article.
- URLs for large images that are supposed to take up the full width within the News Details need to be centered within the Google Document.

The caption works exactly as it does for normal images, which means they are specified in the following paragraph or can be skipped with `[no-caption]`.

Also can these images be linked by specifying `[image-link-to: URL]` in the caption paragraph.

Here is an example of how the Google Document and final result look like:
image
image

### DIFF
--- a/lib/article_json/import/google_doc/html/image_parser.rb
+++ b/lib/article_json/import/google_doc/html/image_parser.rb
@@ -22,25 +22,31 @@ module ArticleJSON
           # The value of the image's `alt` attribute
           # @return [String]
           def alt
+            return '' if image_url?
+
             image_node.attribute('alt')&.value || ''
           end
 
           # The value of the image's `src` attribute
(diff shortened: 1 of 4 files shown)


ifosch:
+1

eporter82:
+1

damiendillon:
+1
Complete with grammar corrections! Good job!

ozgun:
+1 Looks good, thanks!

PR #174, opened 2023-05-17
Update rubocop.yml

damiendillon:
Update outdated `rubocop.yml` to satisfy complaints.

Do not change any settings. Add new categories of
-Security
-Layout
-Naming
and move appropriate cops here.

Rename cops as required and delete those that are redundant.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,27 +22,12 @@ Lint/AssignmentInCondition:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#safe-assignment-in-condition'
   Enabled: true
 
-Lint/BlockAlignment:
-  Description: 'Align block ends correctly.'
-  Enabled: true
-
 Lint/CircularArgumentReference:
   Description: "Don't refer to the keyword argument in the default value."
   Enabled: true
 
-Lint/ConditionPosition:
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -1,6 +1,6 @@
 AllCops:
   DisabledByDefault: true
-  TargetRubyVersion: 2.3
```
dsager: In theory we support 2.3 upwards, see [here](https://github.com/Devex/article_json/blob/master/article_json.gemspec#L26). So I'm not sure if RuboCop should target the latest version. My inclination would be to target the minimal required version.
But we can also increase the minimum required ruby version and drop support for older versions ([Actually all versions <3 have already reached EOL](https://www.ruby-lang.org/en/downloads/branches/)).

damiendillon: Hmm. I can just revert this to 2.3 and if we 
>increase the minimum required ruby version and drop support for older versions ([Actually all versions <3 have already reached EOL](https://www.ruby-lang.org/en/downloads/branches/)).

that can be a separate task.

damiendillon: Reverted to 2.3

ozgun:
:+1:

PR #175, opened 2023-05-17
Rubocop autocorrections

damiendillon:
Enhance code quality with Rubocop autocorrections.

### DIFF
--- a/lib/article_json/elements/base.rb
+++ b/lib/article_json/elements/base.rb
@@ -37,4 +37,3 @@ module ArticleJSON
     end
   end
 end
-
--- a/spec/article_json/configuration_spec.rb
+++ b/spec/article_json/configuration_spec.rb
@@ -121,7 +121,6 @@ describe ArticleJSON::Configuration do
     end
 
     context 'when it has a value' do
-
     end
   end
 end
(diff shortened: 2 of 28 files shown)


damiendillon:
The commit descriptions should perhaps be the commit titles. Apologies for the extra clicks just to read them.

ozgun:
Looks good, thanks :)

agustibr:
👍️ Thanks!

EdLittle:
+1

i thought we liked trailing commas 🤔

damiendillon:
> +1
> 
> i thought we liked trailing commas 🤔

Is it not that we have them in hashes but not arrays?

dsager:
I would actually be in favor of adjusting the rules, so that both Hashes and Arrays have trailing commas, like we do in our main codebase! Any thoughts on this before we move ahead?

ozgun:
My vote is for the consistency. Since our main codebase has trailing commas, I'd say we should follow the same rule here as well.

damiendillon:
Sure, I will remove the trailing commas commit.

damiendillon:
Dropped the commit with the removal of trailing commas in arrays.

PR #176, opened 2023-05-17
WIP BAN-2027 Add Apple News text, paragraph and headlines elements

damiendillon:
Add the
- base
- text
- heading and
- paragraph

elements, along with their spec files.

Each element is returned as a component in the Apple News format with the required keys of `role` and `text` (Later this may be expanded to contain `layout`).

Neither is the key of `format` is applied- as it is not a required field, but we the Apple News format supports HTML and we are returning HTML.

We may wish to consider adding `format` at a later point.

RE `format` the Apple docs note:
> If format is set to html or markdown, neither Additions nor
> InlineTextStyles are supported.
> Default: none
> Possible values: markdown, html, none

### DIFF
--- a/lib/article_json/import/google_doc/html/image_parser.rb
+++ b/lib/article_json/import/google_doc/html/image_parser.rb
@@ -22,25 +22,31 @@ module ArticleJSON
           # The value of the image's `alt` attribute
           # @return [String]
           def alt
+            return '' if image_url?
+
             image_node.attribute('alt')&.value || ''
           end
 
           # The value of the image's `src` attribute
           # @return [String]
           def source_url
+            return @node.inner_text.strip if image_url?
+
             image_node.attribute('src').value
           end
 
           # The node of the actual image
           # @return [Nokogiri::HTML::Node]
           def image_node
-            @node.xpath('.//img').first
+            return @image_node if defined? @image_node
+
+            @image_node = @node.xpath('.//img').first
           end
 
           # Check if the image is floating (left, right or not at all)
           # @return [Symbol]
           def float
-            super if floatable_size?
+            super if image_url? || floatable_size?
           end
 
           # Extracts an href from the tag [image-link-to: url]) if present
@@ -80,6 +86,7 @@ module ArticleJSON
           def href_regexp
             %r{\[image-link-to:\s+(?<url>.*?)\]}
           end
+
           # Check if the image's width can be determined and is less than 500px
           # This is about 3/4 of the google document width...
           # @return [Boolean]
@@ -100,6 +107,13 @@ module ArticleJSON
                 match = image_node.attribute('style').value.match(regex)
                 match['px'].to_i if match && match['px']
               end
+          end
+
+          # When the current node doesn't contain an actual image tag,
+          # we're dealing with an image URL
+          # @return [Boolean]
+          def image_url?
+            image_node.nil?
           end
         end
       end
--- a/lib/article_json/import/google_doc/html/node_analyzer.rb
+++ b/lib/article_json/import/google_doc/html/node_analyzer.rb
@@ -88,7 +88,17 @@ module ArticleJSON
           # @return [Boolean]
           def image?
             return @is_image if defined? @is_image
-            @is_image = node.xpath('.//img').length > 0
+            @is_image = image_url? || node.xpath('.//img').length > 0
+          end
+
+          # Check if the node contains an image URL
+          # @return [Boolean]
+          def image_url?
+            return @is_image_url if defined? @is_image_url
+
+            text = node.inner_text.strip
+            url_regexp = %r{https?:\/\/\S+\.(?:jpg|jpeg|png|gif)}i
+            @is_image_url = !!(url_regexp =~ text)
           end
 
           # Check if the node contains an embedded element
--- a/spec/article_json/import/google_doc/html/image_parser_spec.rb
+++ b/spec/article_json/import/google_doc/html/image_parser_spec.rb
@@ -16,14 +16,19 @@ describe ArticleJSON::Import::GoogleDoc::HTML::ImageParser do
     ArticleJSON::Import::GoogleDoc::HTML::CSSAnalyzer.new(css)
   end
   let(:css) { '' }
-  let(:source_url) { 'foo/bar.jpg' }
+  let(:source_url) { 'https://devex.com/images/bar.jpg' }
   let(:image_fragment) { "<p><span><img src=\"#{source_url}\"></span></p>" }
-  let(:caption_text) { '[image-link-to: http://devex.com] foo' }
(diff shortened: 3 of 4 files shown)


lib/article_json/export/apple_news/elements/paragraph.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Export
+    module AppleNews
+      module Elements
+        class Paragraph < Base
+          # Generate the paragraph node with its containing text elements
+          # @return [Hash]
+          def export
+            {
+              role: 'body',
+              text: text
```
damiendillon: I considered that, as we are returning HTML, that we should possibly tag paragraphs in 'p' tags  in order to maintain consistent HTML, but also thought it may be superfluous. 


Opinions?

```suggestion
              text: "<p>#{text}</p>"
```

damiendillon:
WIP, excuse me.

lib/article_json/export/apple_news/elements/paragraph.rb
agustibr: In my opinion, it makes sense, because of the documentation comment 

> `# Generate the paragraph node with its containing text elements`

lib/article_json/export/apple_news/elements/base.rb
```diff
@@ -0,0 +1,53 @@
...
+    module AppleNews
+      module Elements
+        class Base
+          include ArticleJSON::Export::Common::Elements::Base
+
+          # Export the given element. Dynamically looks up the right
+          # export-element-class, instantiates it and then calls the `#export`
+          # method.
+          # Defaults to an empty string, e.g. if no exporter is specified for
+          # the given type.
```
dsager: The default is `nil`, isn't it?

lib/article_json/export/apple_news/elements/paragraph.rb
dsager: This depends on how Apple defines the format. From the example [here](https://developer.apple.com/documentation/apple_news/body#2973276) I understand that it's not necessary and that `{"role":  "body","text": "some text of the paragraph"}` is a valid paragraph.

Paragraph is what we call this type of element, Apple calls it body. So I don't think it needs to include a `p` tag...

lib/article_json/export/apple_news/elements/text.rb
```diff
@@ -0,0 +1,16 @@
+module ArticleJSON
+  module Export
+    module AppleNews
+      module Elements
+        class Text < Base
+          include ArticleJSON::Export::Common::HTML::Elements::Base
+          include ArticleJSON::Export::Common::HTML::Elements::Text
+
+          def export
```
dsager: I think it'd be worth adding a bit of documentation here, why are we using super and calling to_s on it?

lib/article_json/import/google_doc/html/image_parser.rb
```diff
@@ -22,25 +22,31 @@ def initialize(node:, caption_node:, css_analyzer:)
           # The value of the image's `alt` attribute
           # @return [String]
           def alt
+            return '' if image_url?
```
dsager: I think the changes in this file shouldn't be part of your PR?

lib/article_json/import/google_doc/html/node_analyzer.rb
```diff
@@ -88,7 +88,17 @@ def quote?
           # @return [Boolean]
           def image?
             return @is_image if defined? @is_image
-            @is_image = node.xpath('.//img').length > 0
+            @is_image = image_url? || node.xpath('.//img').length > 0
```
dsager: Changes in this file are also not related to Apple News

spec/article_json/export/apple_news/elements/base_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    context 'when the source element is a heading' do
+      let(:source_element) do
+        ArticleJSON::Elements::Heading.new(content: 'Foo Bar', level: 1)
+      end
+      let(:heading) { { role: 'heading1', text: 'Foo Bar' } }
+      it { should eq heading }
+    end
+
+    context 'when the source element is a paragraph' do
+      let(:source_element) { sample_paragraph }
```
dsager: I think the `sample_paragraph` above can be removed if you specify it directly in this context:
```suggestion
      let(:source_element) { ArticleJSON::Elements::Paragraph.new(content: [sample_text]) }
```

spec/article_json/export/apple_news/elements/heading_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      it { should eq heading }
+    end
+
+    context 'when the heading level is 2' do
+      let(:level) { 2 }
+      let(:heading) { { role: 'heading2', text: 'Foo Bar' } }
+      it { should eq heading }
+    end
+
+    (3..6).each do |i|
```
dsager: Why not use the loop to cover all levels?
```suggestion
    (1..6).each do |i|
```

spec/article_json/import/google_doc/html/image_parser_spec.rb
```diff
@@ -16,14 +16,19 @@
     ArticleJSON::Import::GoogleDoc::HTML::CSSAnalyzer.new(css)
```
dsager: Changes in this file not related to Apple News...

spec/article_json/import/google_doc/html/node_analyzer_spec.rb
```diff
@@ -178,12 +178,58 @@
       it { should be true }
```
dsager: Changes in this file not related to Apple News...

lib/article_json/import/google_doc/html/image_parser.rb
damiendillon: That is correct. This is part the git issue I said I had.

lib/article_json/import/google_doc/html/node_analyzer.rb
damiendillon: That is correct. This is part the git issue I said I had.

spec/article_json/export/apple_news/elements/heading_spec.rb
damiendillon: Good point

spec/article_json/import/google_doc/html/node_analyzer_spec.rb
damiendillon: That is correct. This is part the git issue I said I had.

spec/article_json/import/google_doc/html/image_parser_spec.rb
damiendillon: That is correct. This is part the git issue I said I had.

PR #177, opened 2023-05-18
BAN-2027 Add Apple News text, paragraph and headlines elements

damiendillon:
Add the
- base
- text
- heading and
- paragraph

elements, along with their spec files.

Each element is returned as a component in the Apple News format with the required keys of `role` and `text` (Later this may be expanded to contain `layout`).

Neither is the key of `format` is applied- as it is not a required field, but we the Apple News format supports HTML and we are returning HTML.

We may wish to consider adding `format` at a later point.

RE `format` the Apple docs note:
> If format is set to html or markdown, neither Additions nor
> InlineTextStyles are supported.
> Default: none
> Possible values: markdown, html, none

### DIFF
--- a/lib/article_json.rb
+++ b/lib/article_json.rb
@@ -77,6 +77,11 @@ require_relative 'article_json/export/html/elements/quote'
 require_relative 'article_json/export/html/elements/embed'
 require_relative 'article_json/export/html/exporter'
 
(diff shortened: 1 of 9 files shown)


EdLittle:
+1 lgtm

eporter82:
+1

dsager (approved):

PR #178, opened 2023-05-19
BAN-2027 Add the export to Apple News feature

damiendillon:
Add the
- Apple News Exporter class,
- the script to call it and
- the spec fixture files


spec/article_json/export/apple_news/exporter_spec.rb
```diff
@@ -0,0 +1,13 @@
+describe ArticleJSON::Export::AppleNews::Exporter do
+  subject(:exporter) { described_class.new(elements).to_json }
+
+  let(:json) { File.read('spec/fixtures/apple_news_doc.heading_and_basic_text_parsed.json') }
```
dsager: In the long run this will be the entire reference document, so I would follow the naming scheme of the other files
```suggestion
  let(:json) { File.read('spec/fixtures/reference_document_exported.apple_news.json') }
```

bin/update_reference_document.sh
```diff
@@ -9,6 +9,7 @@ JSON_FILE="spec/fixtures/reference_document_parsed.json"
 HTML_EXPORT_FILE="spec/fixtures/reference_document_exported.html"
 AMP_EXPORT_FILE="spec/fixtures/reference_document_exported.amp.html"
 FACEBOOK_EXPORT_FILE="spec/fixtures/reference_document_exported.facebook.html"
+APPLE_NEWS_EXPORT_FILE="spec/fixtures/heading_and_basic_text_exported.apple_news.html"
```
dsager:
```suggestion
APPLE_NEWS_EXPORT_FILE="spec/fixtures/reference_document_exported.apple_news.json"
```

lib/article_json/export/apple_news/exporter.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Export
+    module AppleNews
+      class Exporter
+        # @param [Array[ArticleJSON::Elements::Base]] elements
+        def initialize(elements)
+          @elements = elements
+        end
+
+        # Return the components section of an Apple News Article as JSON
+        # @return [Hash]
```
dsager: `to_json` will always return a String...
```suggestion
        # @return [String]
```

spec/article_json/export/apple_news/exporter_spec.rb
damiendillon: Amended to:
```
let(:json) { File.read('spec/fixtures/reference_document_parsed.apple_news.json') }
```
(i.e. following the naming scheme- but this is not the exported doc)

bin/update_reference_document.sh
damiendillon: Amended.

lib/article_json/export/apple_news/exporter.rb
damiendillon: Amended

ozgun:
Looks good! Thanks :)

EdLittle:
+1 LGTM

PR #179, opened 2023-05-24
BAN-2029 Add Apple News Image support

damiendillon:
Add image support to the apple news export feature

https://mydevex.atlassian.net/browse/BAN-2029


EgleMekaite:
+1

spec/article_json/export/apple_news/elements/image_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      caption: caption,
+      float: float,
+      alt: alt
+    )
+  end
+  let(:float) { nil }
+  let(:caption) { [ArticleJSON::Elements::Text.new(content: 'Foo Bar')] }
+  let(:alt) { 'Text description' }
+
+  let(:expected_json) { {:URL=>"/foo/bar.jpg", :caption=>"Foo Bar", :role=>"image"} }
```
ozgun: How about `{ URL: "/foo/bar.jpg", caption: "Foo Bar", role: "image" }` ?

ozgun:
Looks good!

When I was checking Apple News API documentation, I came across this: "Image filenames should be properly encoded as URLs." I wonder if we need to encode image URLs. Maybe Google exports them as URL encoded, not sure though.

damiendillon:
> Looks good!
> 
> When I was checking Apple News API documentation, I came across this: "Image filenames should be properly encoded as URLs." I wonder if we need to encode image URLs. Maybe Google exports them as URL encoded, not sure though.

image

The above example url was generated by dragging and dropping an image to a google doc.

damiendillon:
However, maybe we should encode it anyway, as this is a requirement and the input may vary. If you think I should, I am happy to adjust the PR.

spec/article_json/export/apple_news/elements/image_spec.rb
dsager: Yep, we should follow the style we use elsewhere (also using single quotes):
```suggestion
      let(:expected_json) { {URL: '/foo/bar.jpg', caption: 'Foo Bar', role: 'image'} }
```

lib/article_json/export/apple_news/elements/image.rb
```diff
@@ -0,0 +1,19 @@
+module ArticleJSON
+  module Export
+    module AppleNews
+      module Elements
+        class Image < Base
+          # Headline
```
dsager: :eyes:

lib/article_json/export/apple_news/elements/image.rb
```diff
@@ -0,0 +1,19 @@
...
+          # @return [Hash]
+          def export
+            {
+              role: "image",
+              URL: @element.source_url,
```
dsager, line 11: Don't we have to upload images to apple news? Or would that step happen afterwards?

ozgun: It will happen afterwards as far as I remember.

ozgun:
We shouldn't encode already encoded URL.

lib/article_json/export/apple_news/elements/image.rb
damiendillon: 😬

spec/article_json/export/apple_news/elements/image_spec.rb
damiendillon: Amended

lib/article_json/export/apple_news/elements/image.rb
damiendillon: Amended

PR #180, opened 2023-06-07
BAN-2033 Add Apple news 'Videos' export feature

damiendillon:
Use the Apple News Component `EmbedWebVideo` for YouTube, Dailymotion and Vimeo videos.

Also add facebook_post (for Facebook videos) & tweet components.

https://mydevex.atlassian.net/browse/BAN-2033

### DIFF
--- a/lib/article_json/export/apple_news/elements/embed.rb
+++ b/lib/article_json/export/apple_news/elements/embed.rb
@@ -1,0 +1,96 @@
+module ArticleJSON
+  module Export
+    module AppleNews
+      module Elements
+        class Embed < Base
+          # Embed
+          # @return [Hash]
+          def export
+            {
+              role: role,
+              URL: source_url,
+              caption: caption
+            }.compact
+          end
+
+          private
+
+          def role
+            @role ||=
+              case embed_type
+              when :youtube_video, :vimeo_video, :dailymotion_video
+                :embedwebvideo
+              when :facebook_video
+                :facebook_post
+              when :tweet
+                :tweet
+              when :slideshare
+                nil
+              when :soundcloud
+                nil
+              else
+                nil
+              end
+          end
+
+          def source_url
+            case embed_type
+            when :youtube_video
+              build_embeded_youtube_url
+            when :vimeo_video
+              build_embeded_vimeo_url
+            when :dailymotion_video
+              build_embeded_vimeo_url
+            when :facebook_video
+              build_facebook_video_url
+            when :tweet
+              build_twitter_url
+            when :slideshare
+              nil
+            when :soundcloud
+              nil
+            else
+              nil
+            end
+          end
+
+          def caption
+            return nil if role.nil?
+
+            @element.caption.first&.content
+          end
+
+          def build_embeded_youtube_url
+            "https://www.youtube.com/embed/#{embed_id}"
+          end
+
+          def build_embeded_vimeo_url
+            "https://player.vimeo.com/video/#{embed_id}"
+          end
+
+          def build_embeded_dailymotion_url
+            "https://geo.dailymotion.com/player.html?video=#{embed_id}"
+          end
+
+          def build_facebook_video_url
+            username, id = embed_id.to_s.split("/")
+            "https://www.facebook.com/#{username}/videos/#{id}"
+          end
+
+          def build_twitter_url
+            username, id = embed_id.to_s.split("/")
+            "https://twitter.com/#{username}/status/#{id}"
+          end
+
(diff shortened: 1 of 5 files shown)


lib/article_json/export/apple_news/elements/embed.rb
```diff
@@ -0,0 +1,100 @@
...
+
+          def build_facebook_video_url
+            embed_params = embed_id.to_s.split("/")
+            username = embed_params[0]
+            embed_id = embed_params[1]
+            "https://www.facebook.com/#{username}/videos/#{embed_id}"
+          end
+
+          def build_twitter_url
+            embed_params = embed_id.to_s.split("/")
```
ozgun: Not a big deal, but could be simplified like this:
```ruby
username, embed_id = embed_id.to_s.split("/").take(2)
```

lib/article_json/export/apple_news/elements/embed.rb
```diff
@@ -0,0 +1,100 @@
...
+          def role
+            @role ||=
+              case embed_type
+              when :youtube_video, :vimeo_video, :dailymotion_video
+                :embedwebvideo
+              when :facebook_video
+                :facebook_post
+              when :tweet
+                :tweet
+              when :slideshare
```
ozgun, line 27: Do we need `:slideshare` and `:soundcloud` cases here? Or does it mean they will be coming soon?

lib/article_json/export/apple_news/exporter.rb
```diff
@@ -23,6 +23,8 @@ def components
               ArticleJSON::Export::AppleNews::Elements::Base
                 .build(element)
                 &.export
+            end.reject do |hash|
```
ozgun: How about `end.reject { |hash| hash&.empty? }`

ozgun:
I have a couple of minor suggestions, but they don't seem to be blockers.

lib/article_json/export/apple_news/elements/embed.rb
damiendillon: image

lib/article_json/export/apple_news/elements/embed.rb
damiendillon: It does not necessarily mean that they are coming soon but they are elements that we send from article_json and they act as placeholders, showing that they have not been dealt with.

lib/article_json/export/apple_news/exporter.rb
damiendillon: Amened

lib/article_json/export/apple_news/elements/embed.rb
damiendillon: The suggestion should work in theory, but the variable cannot be reused it seems.

Amened without the take `method`.

lib/article_json/export/apple_news/exporter.rb
damiendillon: We actually have nil values too that we want to remove.

damiendillon: Made it a one liner.

damiendillon:
> I have a couple of minor suggestions, but they don't seem to be blockers.

If you are happy with it..

ozgun:
@damiendillon Thanks for the changes :)

damiendillon:
> @damiendillon Thanks for the changes :)

I know there is no push pull on here but if you approve, I think leaving a 'plus one' is clearer for whoever clicks for the merge.

ozgun:
Sure! +1

EdLittle:
+1 LGTM

dsager:
+1 GJ!

lib/article_json/export/apple_news/elements/embed.rb
```diff
@@ -0,0 +1,96 @@
...
+          def build_embeded_vimeo_url
+            "https://player.vimeo.com/video/#{embed_id}"
+          end
+
+          def build_embeded_dailymotion_url
+            "https://geo.dailymotion.com/player.html?video=#{embed_id}"
+          end
+
+          def build_facebook_video_url
+            username, id = embed_id.to_s.split("/")
```
dsager, line 76: you can also tell into how many items you want to split the string: `split("/", 2)`

damiendillon: Ah, good to know. I can add that.

ozgun: Nice, I didn't know about the second parameter!

PR #181, opened 2023-06-12
BAN-2131 Add format key and html value to text output in Apple News

damiendillon:
From manual testing it was noted that it was not enough to just send html, we must also indicate that this is the format we are sending.

This PR adds the format key and value to the text component.

https://mydevex.atlassian.net/browse/BAN-2131


ozgun:
Looks good!

EgleMekaite:
+1 LGTM

dsager:
+1 thanks!

PR #183, opened 2023-06-19
BAN-2137 Correct parsing/display of quotation marks

damiendillon:
Articles in Apple News were showing backslashes before quotation marks, which was because they were being escaped incorrectly.

Escaping the special characters is, in fact unnecessary.

https://mydevex.atlassian.net/browse/BAN-2137


EgleMekaite:
+1

ozgun:
+1 Looks good, thanks!

EdLittle:
+1

PR #185, opened 2023-06-22
BAN-2135 Remove unsupported tags from text in apple news format

EdLittle:
Sometimes we add scripts in the articles to embed some widgets such as audio players to have supplementary content in our news pages. These scripts sometimes have HTML tags that Apple doesn't allow in their news format. We have to omit them from the text.

This is the list of unsupported tags:
  title
  meta
  script
  noscript
  style
  link
  applet
  object
  iframe
  noframes
  form
  select
  option
  optgroup

https://mydevex.atlassian.net/browse/BAN-2135

As an example, in this article, we've been able to attach an audio player to our devex page by inserting a script:

Screenshot 2023-06-23 at 10 59 50

This is what it looks like in the Google Doc:

Screenshot 2023-06-23 at 11 00 07

### DIFF
--- a/lib/article_json/export/apple_news/elements/text.rb
+++ b/lib/article_json/export/apple_news/elements/text.rb
@@ -6,6 +6,10 @@ module ArticleJSON
           include ArticleJSON::Export::Common::HTML::Elements::Base
           include ArticleJSON::Export::Common::HTML::Elements::Text
 
+          UNSUPPORTED_HTML_TAGS = %w[title meta script noscript style link applet object iframe
+            noframes form select option optgroup
+          ].freeze
+
           # A Nokogiri object is returned with`super`, which is is then
           # returned as a either a string or as HTML (when not plain text),
           # both of which are compatible with Apple News format. Takes into
@@ -13,6 +17,23 @@ module ArticleJSON
           # @return [String]
           def export
             super.to_s
+          end
+
+          # @param [String] text
+          def create_text_nodes(text)
+            Nokogiri::HTML.fragment(sanitize_text(text).gsub(/\n/, '<br>')).children
+          end
+
+          # Removes UNSUPPORTED_TAGS from text
+          #
+          # @param [String] text
+          # @return [String]
+          def sanitize_text(text)
+            doc = Nokogiri::HTML.fragment(text)
+            UNSUPPORTED_HTML_TAGS.each do |tag|
+              doc.search(tag).each(&:remove)
+            end
+            doc.inner_html
           end
         end
       end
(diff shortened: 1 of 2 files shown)


EdLittle:
tagging @damiendillon and @ozgun

damiendillon:
+1 LGTM

spec/article_json/export/apple_news/elements/text_spec.rb
```diff
@@ -45,5 +45,14 @@
       let(:output) { '<strong><em>Foo Bar</em></strong>' }
       it { should eq output }
     end
+
+    context 'when the source element has unsupported tags' do
+      let(:content) do
+        'Normal text<script async>Some Script here</script><select><option>OPtion 1</option>' \
```
ozgun: I would add at least one supported HTML tag such as `<h1>Heading 1</h1>` to ensure it is not removed.

EdLittle: Right now, it's not conserving the valid HTML tags but I'm thinking that we should, actually. Im going to change it so that it only removes the unsupported tags

EdLittle:
Latest Update: Conserve supported HTML tags

spec/article_json/export/apple_news/elements/text_spec.rb
ozgun: Does the `inner_html` method also work with nested tags? For example if the input is something like this:

```html
Normal text
<script async>Some Script here</script>
<select><option>OPtion 1</option></select>
<form>form content</form>
<p><em>text</em></p>
<form><a href="#">Link</a></form>
```

lib/article_json/export/apple_news/elements/text.rb
```diff
@@ -14,6 +18,23 @@ class Text < Base
           def export
             super.to_s
           end
+
+          # @param [Nokogiri::XML::NodeSet] text
+          def create_text_nodes(text)
+            Nokogiri::HTML.fragment(sanitize_text(text).gsub(/\n/, '<br>')).children
```
dsager, line 24: This looks a bit weird to me, why are there line breaks in the text? The Google Doc HTML export should already convert all those to `<br/>` tags, unless I'm missing something :thinking:

dsager: Never mind, I just saw that the overwritten method does this too...

lib/article_json/export/apple_news/elements/text.rb
```diff
@@ -14,6 +18,23 @@ class Text < Base
...
+          end
+
+          # Removes UNSUPPORTED_TAGS from text
+          #
+          # @param [String] text
```
dsager, line 29: According to the other methods, `text` is a `Nokogiri::XML::NodeSet`.
Shouldn't we be able to iterate over it directly, without creating another `Nokogiri::HTML.fragment`?

dsager: OK, Apparently the method doc here is wrong: https://github.com/Devex/article_json/blob/master/lib/article_json/export/common/html/elements/text.rb#L46

PR #186, opened 2023-06-28
BAN-2160 Show image captions on Apple News

damiendillon:
The article_json gem is outputting JSON with which Apple News should be showing image captions in articles.
However these captions were not showing in any form.

This PR makes the necessary adjustments so that captions show in Apple News articles.

A follow-up PR with deal with the same issue for captions not showing for videos.

https://mydevex.atlassian.net/browse/BAN-2160

### DIFF
--- a/lib/article_json/export/apple_news/elements/image.rb
+++ b/lib/article_json/export/apple_news/elements/image.rb
@@ -3,14 +3,38 @@ module ArticleJSON
     module AppleNews
       module Elements
         class Image < Base
+          # Image | Image, Caption
+          # @return [Hash, Array<Hash>]
+          def export
+            caption_text.nil? ? image : [image, caption]
+          end
+
+          private
           # Image
           # @return [Hash]
-          def export
+          def image
             {
-              role: "image",
+              role: 'image',
               URL: @element.source_url,
-              caption: @element.caption.first&.content
+              caption: caption_text,
             }.compact
+          end
+
+          # Caption
+          # @return [Hash]
+          def caption
+            {
+              role: 'caption',
+              text: caption_text,
+              layout: 'captionLayout',
+              textStyle: 'captionStyle',
+            }
+          end
+
+          # Caption Text
+          # @return [String]
+          def caption_text
+            @element.caption.first&.content
           end
         end
       end
--- a/lib/article_json/export/apple_news/exporter.rb
+++ b/lib/article_json/export/apple_news/exporter.rb
@@ -8,15 +8,21 @@ module ArticleJSON
         end
 
         # Return the components section of an Apple News Article as JSON
+        #
+        # Images and EmbededVideos are nested in an array with the components
+        # array when they contain captions. As Apple News skips over these
+        # nested arrays, we must flatten the array.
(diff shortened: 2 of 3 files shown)


lib/article_json/export/apple_news/elements/image.rb
```diff
@@ -3,14 +3,37 @@ module Export
     module AppleNews
       module Elements
         class Image < Base
+          # Image | Image, Caption
+          # @return [Hash] | [Hash], [Hash]
+          def export
+            caption_text.nil? ? image : [image, caption].join(',')
```
dsager: This looks weird, you're converting the hashes into a string here. Wouldn't it make more sense to return the array of hashes here?
```suggestion
            caption_text.nil? ? image : [image, caption]
```
BTW, the YARD notation for this is would then be `@return [Hash, Array<Hash>]`

spec/article_json/export/apple_news/elements/image_spec.rb
```diff
@@ -13,29 +13,43 @@
...
+        role: 'image',
+        URL: '/foo/bar.jpg',
+      },
+      {
+        role: 'caption',
+        text: 'Foo Bar',
+        layout: 'captionLayout',
+        textStyle: 'captionStyle',
+      }
+    ].join(',')
```
dsager: This won't produce JSON, but a string of concatenated `Hash#to_s`...

lib/article_json/export/apple_news/elements/image.rb
damiendillon: So, I was trying to return the JSON contents (without the array) but I realise now that I can/should just call ```.flatten``` on the components array.

spec/article_json/export/apple_news/elements/image_spec.rb
damiendillon: Ditto.

damiendillon:
Pushed a change to return JSON in an array and then flatten the whole components array as if the JSON is sitting in a nested array it is skipped over by Apple News.

lib/article_json/export/apple_news/exporter.rb
```diff
@@ -10,7 +10,7 @@ def initialize(elements)
         # Return the components section of an Apple News Article as JSON
         # @return [String]
         def to_json
-          { components: components }.to_json
+          { components: components.flatten }.to_json
```
dsager, line 18: Maybe a comment why we're calling `flatten` here would be helpful?

dsager (approved):

lib/article_json/export/apple_news/exporter.rb
damiendillon: I considered that but thought I might be over-explaining. I will happily add it now.

damiendillon: Added a comment to explain `.flatten` and correct the YARD `@return` for the `components` method to reflect that it returns an array rather than a string.

damiendillon:
image

EdLittle:
+1

PR #189, opened 2023-07-06
BAN-2178 Add support for lists to Apple News

damiendillon:
Add support for numbered and unnumbered list items. 
https://mydevex.atlassian.net/browse/BAN-2178


lib/article_json/export/apple_news/elements/list.rb
```diff
@@ -0,0 +1,71 @@
+module ArticleJSON
+  module Export
+    module AppleNews
+      module Elements
+        class List < Base
+          include ArticleJSON::Export::Common::HTML::Elements::Base
+          include ArticleJSON::Export::Common::HTML::Elements::Text
```
dsager: Are both of these needed to have access to `self.class.exporter_by_type`?

lib/article_json/export/apple_news/elements/list.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          # Each list item should be wrapped in <li></li>
+          #
+          # @return [String]
+          def list
+            @element.content.map do |paragraph_element|
+              line_item = paragraph_element.content.map do |text_element|
+                text_exporter.new(text_element).export
+              end
+              .join
```
ozgun: I would do `end.join` to be consistent with the line 50.

ozgun:
In general it looks good to me, thanks! But, maybe worth to check @dsager 's comment above.

lib/article_json/export/apple_news/elements/list.rb
damiendillon: I think they are actually possibly both redundant. Tests pass without them. I will do a manual test to confirm.

damiendillon: Manual test passed. I will remove them now.

lib/article_json/export/apple_news/elements/list.rb
damiendillon: Amended

lib/article_json/export/apple_news/elements/list.rb
damiendillon: Amended, thanks.

PR #191, opened 2023-07-13
BAN-2160 Show links in captions

damiendillon:
Add support for links in captions of images and embed features.

https://mydevex.atlassian.net/browse/BAN-2160

### DIFF
--- a/lib/article_json/export/apple_news/elements/image.rb
+++ b/lib/article_json/export/apple_news/elements/image.rb
@@ -26,15 +26,30 @@ module ArticleJSON
             {
               role: 'caption',
               text: caption_text,
+              format: 'html',
               layout: 'captionLayout',
               textStyle: 'captionStyle',
             }
           end
 
+          # Get the exporter class for text elements
+          # @return [ArticleJSON::Export::Common::HTML::Elements::Base]
+          def text_exporter
+            self.class.exporter_by_type(:text)
+          end
+
           # Caption Text
           # @return [String]
           def caption_text
-            @element.caption.first&.content
+            text.empty? ? nil : text
+          end
+
+          # @return [String]
+          def text
+            @element.caption.map do |child_element|
+              text_exporter.new(child_element)
+                .export
+            end.join
           end
         end
       end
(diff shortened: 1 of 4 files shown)


dsager:
LGTM.
The code for captions looks pretty similar, I wonder if it makes sense to share it between `Embed` and `Image`?

spec/article_json/export/apple_news/elements/image_spec.rb
```diff
@@ -33,17 +35,29 @@
...
+        end
+
         it { should eq expected_json }
       end
 
-      context 'and when the image is floating on the right' do
-        let(:float) { :right }
+      context 'and the caption has got a link' do
+        let(:caption_text) do
+          "This has a link: <a href=\"http://en.wikipedia.org/wiki/Lorem_ipsum\">Lorem ipsum</a>"
```
ozgun, line 59: I'd use single quotes:
```ruby
'This has a link: <a href="http://en.wikipedia.org/wiki/Lorem_ipsum">Lorem ipsum</a>'
```

lib/article_json/export/apple_news/elements/image.rb
```diff
@@ -26,15 +26,30 @@ def caption
...
           # @return [String]
           def caption_text
-            @element.caption.first&.content
+            text.empty? ? nil : text
+          end
+
+          # @return [String]
+          def text
+            @element.caption.map do |child_element|
+              text_exporter.new(child_element)
```
ozgun, line 50: This looks like a one liner to me :)

ozgun:
+1 Looks good, I've just added a few minor comments, thanks!

ozgun:
> The code for captions looks pretty similar, I wonder if it makes sense to share it between Embed and Image?

Good idea, but if it's gonna take time, I'd say let's merge this now and refactor later because there is already a ticket waiting for this fix.

damiendillon:
> LGTM.
> The code for captions looks pretty similar, I wonder if it makes sense to share it between `Embed` and `Image`?

Definitely does. I am pretty confident I have almost the same code in one other place too, but like Oz said, we wanted it to quickly as it is a reopened ticket and this was not really in the original scope as I had not factored links so we want to close it. 

I will give all the article json stuff a refactoring/ cleanup with Rubocop soon.

PR #193, opened 2023-07-13
Update reference document output

damiendillon:
The changes in the Apple News exporter mean that the output from the exporter needs to be different.

This PR updates the reference document output.


dsager:
LGTM

PR #202, opened 2024-01-22
Use checkout v4 for actions

damiendillon:
To avoid this warning that we are getting:
>The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2.
>For more info:
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


dsager (approved):
Thanks!

ozgun (commented):
Thanks!

PR #205, opened 2024-01-31
Ignore `source_url` when checking document export

damiendillon:
The `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison.

### DIFF
--- a/bin/check_google_doc_export.rb
+++ b/bin/check_google_doc_export.rb
@@ -16,12 +16,26 @@ require 'uri'
 require_relative '../lib/article_json'
 
 doc_id = '1E4lncZE2jDkbE34eDyYQmXKA9O26BHUiwguz4S9qyE8'
-url = "https://docs.google.com/feeds/download/documents/export/Export?id=#{doc_id}&exportFormat=html"
-expected_document = JSON.parse(File.read('spec/fixtures/reference_document_parsed.json'))
+url =
+  "https://docs.google.com/feeds/download/documents/export/Export?id=#{doc_id}&exportFormat=html"
+parsed_exported_doc = JSON.parse(
+  ArticleJSON::Article
+    .from_google_doc_html(Net::HTTP.get(URI.parse(url)))
+    .to_json
+)
+parsed_expected_doc = JSON.parse(File.read('spec/fixtures/reference_document_parsed.json'))
 
-exported_doc = Net::HTTP.get(URI.parse(url))
-document = JSON.parse(ArticleJSON::Article.from_google_doc_html(exported_doc).to_json)
+# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
+def nullify_source_url(hash)
+  hash['content'].each { |element| element['source_url'] = nil if element['source_url'] }
+  hash
+end
 
-raise StandardError, "Google doc export doesn't work as expected" if document != expected_document
+parsed_exported_doc = nullify_source_url(parsed_exported_doc)
+parsed_expected_doc = nullify_source_url(parsed_expected_doc)
+
+if parsed_exported_doc != parsed_expected_doc
+  raise StandardError, "Google doc export doesn't work as expected"
+end
 
 puts 'Google doc export worked as expected'


bin/check_google_doc_export.rb
```diff
@@ -16,12 +16,30 @@
...
 
 exported_doc = Net::HTTP.get(URI.parse(url))
 document = JSON.parse(ArticleJSON::Article.from_google_doc_html(exported_doc).to_json)
 
+# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
+def nullify_source_url(hash)
+  content = hash['content']
+  return unless content.is_a?(Array)
+
+  content.each do |key, value|
```
ozgun: When the `content` is an array, then the `value` is always `nil`, isn't it?

bin/check_google_doc_export.rb
```diff
@@ -16,12 +16,30 @@
...
+  "https://docs.google.com/feeds/download/documents/export/Export?id=#{doc_id}&exportFormat=html"
 expected_document = JSON.parse(File.read('spec/fixtures/reference_document_parsed.json'))
 
 exported_doc = Net::HTTP.get(URI.parse(url))
 document = JSON.parse(ArticleJSON::Article.from_google_doc_html(exported_doc).to_json)
 
+# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
+def nullify_source_url(hash)
+  content = hash['content']
+  return unless content.is_a?(Array)
```
ozgun: When we return `nil` here, in the assignment `expected_document = nullify_source_url(expected_document)`, we'll end up with a nil `expected_document`, is this what we want?

bin/check_google_doc_export.rb
damiendillon: Amended, thanks.

bin/check_google_doc_export.rb
damiendillon: From printing the value of the `expected_document` I never did return `nil` here, but you are right, I should have been returning the `content`, rather than `nil`.

Anyway, it has been amended to a one-liner that does not include this early return. Thanks.

bin/check_google_doc_export.rb
```diff
@@ -16,12 +16,21 @@
...
 
 exported_doc = Net::HTTP.get(URI.parse(url))
 document = JSON.parse(ArticleJSON::Article.from_google_doc_html(exported_doc).to_json)
 
+# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
+def nullify_source_url(hash)
+  hash['content'].each { |key, _value| key['source_url'] = nil if key['source_url'] }
+end
+
+expected_document = nullify_source_url(expected_document)
```
ozgun: The `nullify_source_url(expected_document)` call returns `expected_document['content']` not `expected_document` itself as far as I understand, is this what we need here? Because the variables names here are a bit confusing to me.

damiendillon: Now it returns the hash, thanks.

The difference was just this:
```
{"article_json_version"=>"0.4.0", "content"=>[content]}
```
versus
```
content
```
but now it exactly represents the variables.

bin/check_google_doc_export.rb
```diff
@@ -16,12 +16,22 @@
 require_relative '../lib/article_json'
 
 doc_id = '1E4lncZE2jDkbE34eDyYQmXKA9O26BHUiwguz4S9qyE8'
-url = "https://docs.google.com/feeds/download/documents/export/Export?id=#{doc_id}&exportFormat=html"
+url =
+  "https://docs.google.com/feeds/download/documents/export/Export?id=#{doc_id}&exportFormat=html"
 expected_document = JSON.parse(File.read('spec/fixtures/reference_document_parsed.json'))
 
 exported_doc = Net::HTTP.get(URI.parse(url))
 document = JSON.parse(ArticleJSON::Article.from_google_doc_html(exported_doc).to_json)
```
dsager: Not your change, but is there a reason for serializing and deserializing to & from JSON? Could that work too?
```suggestion
document = ArticleJSON::Article.from_google_doc_html(exported_doc).to_h
```

bin/check_google_doc_export.rb
```diff
@@ -16,12 +16,22 @@
...
 
+# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
+def nullify_source_url(hash)
+  hash['content'].each { |key, _value| key['source_url'] = nil if key['source_url'] }
```
dsager: I think `key` isn't the right name to use here. Assuming that `hash['content']` is an array of hashes (which are the objects we call elements), this might be a clearer alternative:
```suggestion
  hash['content'].each { |element| element['source_url'] = nil if element['source_url'] }
```

bin/check_google_doc_export.rb
damiendillon: I will try it

bin/check_google_doc_export.rb
damiendillon: This is cleaner and I have amended it to this. Thanks.

bin/check_google_doc_export.rb
damiendillon: So the 'good reason' is to compare like with like.

damiendillon: Alternatively I could move the parsing into the `nullify` method but it is not really any better placed there IMO, at least not without the method being renamed>

```ruby
doc_export = ArticleJSON::Article.from_google_doc_html(exported_doc).to_json

expected_document = File.read('spec/fixtures/reference_document_parsed.json')

# `source_url` (for hosted images) is dynamic, so we need to remove it from the comparison`
def nullify_source_url(hash)
  hash = JSON.parse(hash)
  hash['content'].each { |element| element['source_url'] = nil if element['source_url'] }
  hash
end
```

dsager: > So the 'good reason' is to compare like with like.

I don't get what you mean. Unless I'm missing something, we have a hash, convert it to json (`.to_json`), then convert it back into a hash again (`JSON.parse`).
My point was to ideally work with the hash directly. Splitting the steps up would obfuscate it even further, so I don't see a good reason for that...

dsager: I guess one reason could be that the parsed JSON would give you string keys, while the internal hash has Symbol keys? But then working with `transform_keys` might be clearer?

dsager: Would this work?
```ruby
exported_article = nullify_source_url(
  ArticleJSON::Article.from_google_doc_html(
    Net::HTTP.get(URI.parse(url))
  )
)
expected_article = nullify_source_url(
  ArticleJSON::Article.from_json(
    File.read('spec/fixtures/reference_document_parsed.json')
  )
)

exported_article == expected_article

dsager (approved):
Approving, as the open conversation is unrelated to your changes...

bin/check_google_doc_export.rb
damiendillon: > I guess one reason could be that the parsed JSON would give you string keys, while the internal hash has Symbol keys?

That was what I meant with 'like' with 'like'. Sorry for not being more explicit

There is deep nesting and do not have Rails `hash_with_indifferent` nor `deep_transform_keys` available to us.

damiendillon:
I tidied up the variable names a little

dsager (approved):

ozgun (approved):
Thanks!

PR #209, opened 2024-11-07
Update rubocop.yml

damiendillon:
Many of the cops have changed since this was last updated. 

This updates them and also adds new cops aligning it with Devex BE guild agreements.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,1102 +1,259 @@
 AllCops:
-  DisabledByDefault: true
+  RubyInterpreters:
+    - ruby
+  Exclude:
+    - 'bin/**/*'
+    - 'script/**/*'
+    - 'tmp/**/*'
+    - '.git/**/*'
+
+  EnabledByDefault: true
+  NewCops: enable
   TargetRubyVersion: 3.2
 
-#################### Lint ################################
 
-Lint/AmbiguousOperator:
-  Description: >-
-                 Checks for ambiguous operators in the first argument of a
-                 method invocation without parentheses.
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#parens-as-args'
-  Enabled: true
-
-Lint/AmbiguousRegexpLiteral:
-  Description: >-
-                 Checks for ambiguous regexp literals in the first argument of
-                 a method invocation without parenthesis.
-  Enabled: true
-
-Lint/AssignmentInCondition:
-  Description: "Don't use assignment in conditions."
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#safe-assignment-in-condition'
-  Enabled: true
-
-Lint/CircularArgumentReference:
-  Description: "Don't refer to the keyword argument in the default value."
-  Enabled: true
-
-Lint/Debugger:
-  Description: 'Check for debugger calls.'
-  Enabled: true
-
-Lint/DeprecatedClassMethods:
-  Description: 'Check for deprecated class method calls.'
-  Enabled: true
-
-Lint/DuplicateMethods:
-  Description: 'Check for duplicate methods calls.'
-  Enabled: true
-
-Lint/EachWithObjectArgument:
-  Description: 'Check for immutable argument given to each_with_object.'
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -1,1102 +1,394 @@
...
-  Description: >-
-                  Use `tr` instead of `gsub` when you are replacing the same
-                  number of characters. Use `delete` instead of `gsub` when
-                  you are deleting characters.
-  Reference: 'https://github.com/JuanitoFatas/fast-ruby#stringgsub-vs-stringtr-code'
-  Enabled: true
-
-##################### Rails ##################################
+#
+# Rails Cops
```
dsager: We should not include any rails cops, as the `gem` does not depend on rails and can be used in non-rails environments.

.rubocop.yml
```diff
@@ -1,1102 +1,394 @@
...
-  Enabled: false
-
+  EnforcedStyle: exploded
+Style/RedundantArrayConstructor:
+  # Excluded so that Grape param declarations, such as:
+  #   type: Array[Integer]/Array[String]/type: Array[Symbol]
+  # do not have to be disabled individually inline, as they would otherwise trigger this cop
+  Exclude:
+    - 'app/controllers/**/*'
+    - 'app/helpers/**/*'
```
dsager: As this is a gem, there is no `app` folder

.rubocop.yml
```diff
@@ -1,1102 +1,394 @@
...
-  Enabled: false
-
-Style/GuardClause:
-  Description: 'Check for conditionals that can be replaced with guard clauses'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-nested-conditionals'
-  Enabled: false
-
+  # Built-in global variables are allowed by default.
+  AllowedVariables:
+    - $flipper
```
dsager: the gem does not have a dependency on Flipper, so I'd remove this

.rubocop.yml
```diff
@@ -1,1102 +1,394 @@
...
-
-Style/EmptyLiteral:
-  Description: 'Prefer literals to Array.new/Hash.new/String.new.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#literal-array-hash'
+  # do not require top level docs for classes/modules
+  Enabled: false
+Style/DocumentationMethod:
+  Exclude:
+    # do not require method documentation in migration files
+    - 'db/migrations/*.rb'
```
dsager: the gem doesn't have a `db` folder.

.rubocop.yml
```diff
@@ -1,1102 +1,394 @@
...
+  EnforcedStyle: to_not
+RSpec/SpecFilePathFormat:
+  CustomTransform:
+    RuboCop: rubocop
+    RSpec: rspec
+    OAuthService: oauth_service
+    OAuthController: oauth_controller
+    API: ''
+  Exclude:
+    - 'spec/lib/devex/api/**/*'
```
dsager: I think this can be removed

damiendillon:
I can remove everything but I left it all as was aligning it with the idea that we would have a shared rubocop file that we spoke about a long while back.

dsager:
> I can remove everything but I left it all as was aligning it with the idea that we would have a shared rubocop file that we spoke about a long while back.

While this makes sense for our internal applications, I think open source gems are a different use case and the rubocop configuration should be tailored for it.

.rubocop.yml
damiendillon: Removed.

.rubocop.yml
damiendillon: Removed.

.rubocop.yml
damiendillon: Removed.

.rubocop.yml
damiendillon: Removed.

.rubocop.yml
damiendillon: Removed.

ozgun (approved):

dsager (approved):

PR #212, opened 2026-03-06
Add `transform_api_response` hook and improve YouTube thumbnail resolution

agustibr:
This pull request  adds a `transform_api_response` hook to the OEmbed resolver base class to allow provider-specific transformations of API responses.

The YouTube resolver now upgrades thumbnails from `hqdefault.jpg` to `maxresdefault.jpg` to use the highest available thumbnail resolution.

The pull request also includes a version bump and updates to test fixtures.


damiendillon (approved):

ozgun (approved):
Looks good!

lib/article_json/utils/o_embed_resolver/youtube_video.rb
```diff
@@ -19,6 +19,19 @@ def oembed_url
...
+        protected
+
+        # @param [Hash] data
+        # @return [Hash]
+        def transform_api_response(data)
+          return data unless data&.dig(:thumbnail_url)
+          
+          # Replace the default YouTube thumbnail with the max resolution version
+          data.merge(
+            thumbnail_url: data[:thumbnail_url].sub(%r{hqdefault\.jpg$}, 'maxresdefault.jpg')
```
bbrassart, line 32: One thing to take into account: it appears that maxresdefault is not always present. Most of the time it is but maybe not always. Anyways, maybe we can try to run it like that and see if we found cases of missing maxresdefault

damiendillon:
BTW this repo does not auto-merge I do not think.

lib/article_json/utils/o_embed_resolver/youtube_video.rb
agustibr: Looks like for old videos, `maxresdefault` might not be available, so they would just use the default thumbnail. We could try this approach and see if any cases break.

